PR #1560, opened 2024-10-05
Draft: adjust goals and homotopy parameter during optimization.

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 11:51

This branch introduces ways to adjust the list of goals during optimization and making it depend on for example the homotopy parameter. It also enables the homotopy parameter theta to depend on time.

This branch seems unnecessary, since there are simpler ways to implement these properties without adding additional features to rtc-tools; see, for example, the comments in this MR and https://gitlab.com/deltares/bos-rijnland/-/merge_requests/42.

### DIFF
--- a/src/rtctools/optimization/goal_programming_mixin.py
+++ b/src/rtctools/optimization/goal_programming_mixin.py
@@ -351,8 +351,9 @@ class GoalProgrammingMixin(_GoalProgrammingMixinBase):
         if goal.has_target_bounds:
             # We use a violation variable formulation, with the violation
             # variables epsilon bounded between 0 and 1.
-            m, M = np.full_like(epsilon, -np.inf, dtype=np.float64), np.full_like(
-                epsilon, np.inf, dtype=np.float64
+            m, M = (
+                np.full_like(epsilon, -np.inf, dtype=np.float64),
+                np.full_like(epsilon, np.inf, dtype=np.float64),
             )
 
             # A function range does not have to be specified for critical
--- a/src/rtctools/optimization/goal_programming_mixin.py
+++ b/src/rtctools/optimization/goal_programming_mixin.py
@@ -625,7 +625,7 @@ class GoalProgrammingMixin(_GoalProgrammingMixinBase):
             self.pre()
 
         # Group goals into subproblems
-        subproblems = []
+        self.__subproblems = []
         goals = self.goals()
         path_goals = self.path_goals()
 
@@ -647,7 +647,7 @@ class GoalProgrammingMixin(_GoalProgrammingMixinBase):
         }
 
         for priority in sorted(priorities):
-            subproblems.append(
+            self.__subproblems.append(
                 (
                     priority,
                     [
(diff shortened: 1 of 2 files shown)


SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 11:51

requested review from @SGeeversAtVortech

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 13:03

Commented on [src/rtctools/optimization/goal_programming_mixin.py line 629](https://github.com/Deltares/rtc-tools/compare/7f078d793396303774d150ae5ebabb7534420ace..a3046720cf5d055e84c16021b64c8f77ebded273#diff-7bfb16ea2a46a61b6bb3da05e71ed2d0R629)



Why is subproblems replaced by self.__subproblems? It doesn't seem to be accessed outside of this method and it is a private variable, so you should not access it from outside this class.

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 13:24

There seems to be a new feature for setting theta as a timeseries, but this doesn't seem to be tested. Please add a unit test for this feature.

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 13:30

Commented on [src/rtctools/optimization/homotopy_mixin.py line 65](https://github.com/Deltares/rtc-tools/compare/45ebfa136c1048df28769f54dd55a34fc394a65f..8f49446c794f952aba8cfb914e3bd1bdde49788d#diff-3737bc68384e801e8dcf29fee1633ec2R65)



Why not use self.__theta here? That should be correct.

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 13:55

Commented on [src/rtctools/optimization/homotopy_mixin.py line 40](https://github.com/Deltares/rtc-tools/compare/7f078d793396303774d150ae5ebabb7534420ace..a3046720cf5d055e84c16021b64c8f77ebded273#diff-3737bc68384e801e8dcf29fee1633ec2R40)



This syntax seems wrong. We're comparing a boolean with a float here. Please also check other comparisons involving .any().

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 17:01

Commented on [src/rtctools/optimization/homotopy_mixin.py line 58](https://github.com/Deltares/rtc-tools/compare/7f078d793396303774d150ae5ebabb7534420ace..a3046720cf5d055e84c16021b64c8f77ebded273#diff-3737bc68384e801e8dcf29fee1633ec2R58)



Can users set this parameter? How should they know?

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 17:07

It seems that in some applications, the private variable __subproblems are accessed and adjusted based on the homotopy parameter theta.
Instead, we could adjust the goals and pathgoals method
to select goals of certain priorities based on the theta parameter.

The goals method could look something like this:

```python
def goals(self, ...):
    goals = ...  # usual calculation of goals
    theta = self.parameters()["theta"]
    priorities = ... # some logic based on theta
    goals = [goal for goal in goals if goal.priority in priorities]
    return goals
```

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on May 16, 2024, 17:21

It seems that here, we try to add a feature for setting a homotopy parameter theta
that can depend on time in a very specific way.
Instead, the user could do the following:
call the homotopy parameter `parameter Real theta_`
and add an input `input Real theta(fixed=true)` to the Modelica file.
Then overwrite the `homotopy_options` method to set `homotopy_parameter` to `theta_`.
Then extend the priority_started method to calculate theta.
This method could look as follows.

```python
def priority_started()
    ... # usual stuff
    theta_ = self.parameters()["theta_"]
    times = ...
    theta = ... # calculate theta
    self.set_timeseries("theta", times, theta)
```

PR #1569, opened 2024-10-07
Do not store timeseries index as timeseries values

SGeeversAtVortech:
In GitLab by @vreeken on Apr 21, 2017, 16:55


SGeeversAtVortech:
In GitLab by @vreeken on Apr 21, 2017, 16:58

I think there is something to be said for also converting directly to floats, i.e. the "CSVMixin: Ensure timeseries are always of dtype np.float64" commit. So it's either this one, or re-reverting that one.

SGeeversAtVortech:
In GitLab by @baayen on Apr 24, 2017, 09:12

closed

PR #1574, opened 2024-10-07
WIP: Implement a mixin to smooth model dynamics

### LINKED ISSUE #1047: Scaling the objective by the number of goals
In GitLab by @TPiovesan on Jun 18, 2018, 12:11

At each priority the objective function that the solver has to minimize is equal to the sum of the objectives of each goals divided by the number of goals:

```python 
def objective(self, ensemble_member):
    if len(self.__subproblem_objectives) > 0:
        acc_objective = ca.sum1(ca.vertcat(*[o(self, ensemble_member) for o in self.__subproblem_objectives]))
        return acc_objective / len(self.__subproblem_objectives)
``` 
(Similarly for the `path_objective`.)

This division plays against the tolerance of the solver. Indeed, if `len(self.__subproblem_objectives) = 100.000` and `solver_tolerance = 1e-8`, the objective function will be in practice have been optimized up to a tolerance of `1e-3`. (Here we assume that the solver won't perform any internal scaling.)

This lost of significant digits of the objective function is particularly problematic at the next priority level, when soft constraints are transformed into hard ones. An epsilon that was considered optimal by the solver may not be accurate enough when constructing the corresponding hard constraint, thus leading to an infeasible problem.

To be discussed:
* what are the advantages of this scaling? 
* if we opt to keep the scaling, what actions should we take to prevent infeasibilities issues?

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 11, 2018, 19:51

In every model ever built, we observe that the control variables induce
undesirably jumpy and pointed results into the model dynamics. We
usually try to resolve this by adding rate-of-change goals ad-hoc onto
the most misbehaved states. This new mixin:

1. Provides a standardized approach to smoothing dynamics
2. Avoids ad-hoc boilerplate der() goals
3. Makes it easy for new users to get agreeable results out of RTC

Closes #1047


SGeeversAtVortech:
In GitLab by @vreeken on Jul 12, 2018, 24:07

How often does it occur that freedom remains in the solution? I would assume that most optimization problems have a minimization goal at the end, which should leave relatively little (if at any) freedom.

Otherwise, I have no issue with adding something like this for cases where such minimization goals do not exist (or freedom remains). What's missing/wrong though is:
- A documentation page (autoclass/module should suffice).
- Explaining/referencing this mixin (and its effect on the results) in the cascade example documentation.
- The function nominal should be used for scaling, not the function range. See 5ac6be99c978ccbceb08247878af3105fb508d09

SGeeversAtVortech:
In GitLab by @baayen on Jul 12, 2018, 08:29

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 17](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..7466a945ed79e8c17e465b8f5878d01444975781#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR17)



Shouldn't this be `min_timestep` so to obtain the largest possible `abs_max_der`?

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 12, 2018, 16:04

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 17](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..7466a945ed79e8c17e465b8f5878d01444975781#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR17)



correct. thanks.

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 12, 2018, 16:04

resolved all discussions

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 12, 2018, 16:37

Not all models are minimized- in some models where pumps dominate they are, but in many models weirs, turbines, or orifices are the dominant control structures. With these structures it is usually preferred to minimize the derivative of the flow rather than the actual flow. And even with a model where the only control variables are pumps, rapidly changing flow rates are also not desirable. It usually best to minimize total cost, leaving flexibility to minimize the derivatives at the end.

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 12, 2018, 20:26

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 17](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..7466a945ed79e8c17e465b8f5878d01444975781#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR17)



changed this line in [version 2 of the diff](https://gitlab.com/deltares/rtc-tools/merge_requests/143/diffs?diff_id=19578278&start_sha=7466a945ed79e8c17e465b8f5878d01444975781#31b891637add97231e53b637a29eef553c1bbf4c_17_17)

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 12, 2018, 20:26

added 1 commit

<ul><li>cd7f87c1 - Implement a mixin to smooth model dynamics</li></ul>

[Compare with previous version](https://gitlab.com/deltares/rtc-tools/merge_requests/143/diffs?diff_id=19578278&start_sha=7466a945ed79e8c17e465b8f5878d01444975781)

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 04:55

added 1 commit

<ul><li>b3066a57 - Implement a mixin to smooth model dynamics</li></ul>

[Compare with previous version](https://gitlab.com/deltares/rtc-tools/merge_requests/143/diffs?diff_id=19588514&start_sha=cd7f87c1e1d8c36e34a2d96e1973dcb2269e1dc6)

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 04:58

@vreeken have a look now- I think I addressed all your concerns.

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 09:59

Commented on [doc/examples/optimization/goal_programming.rst line 120](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-1bd69329a87bb2c391702e90793a4af6R120)



A welcome fix, but unrelated to the commit.

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 10:02

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 43](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR43)



What do you mean with aggressive. From what I can tell it only scales the objective? I agree that the user should be able to influence how the objective is scaled to avoid scaling issues, but it should be clear that that's what it does.

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 10:05

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 57](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR57)



Should we allow users to pass either a list, or a dict of state -> weight? That way users can prefer to minimize derivatives of one state over another.

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 10:07

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 61](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR61)



Wouldn't it be easier to have one single goal which returns (a weighted) sum of all the states?

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 10:09

It could be desirable to also have this Mixin work for fully linear models. That would mean that the order of the minimization goal can be set to 1 (indirectly), and an additional helper variable is needed for the absolute value of a derivative state.

It would also be possible to get a much better nominal (especially important with order=2 I think) by using results of previous goals to get an estimate. This is not possible using individual goals (as then the weighting/relative importance would change), but would be possible for the sum of all der()'s in a single goal.

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 14:41

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 57](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR57)



If they want that fine level of control, then they should implement their own goals. I would say that is outside of the scope of this mixin- we would over-complicate the API. Also, I generally find that using weights is only useful when the model is linear.

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 14:42

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 57](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR57)



On the other hand, I think the initial state estimation mixin does have a way to pass in weights, right?

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 14:44

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 61](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR61)



easier? not sure- this was easy to implement this way. Sorting out the scaling was really simple.

SGeeversAtVortech:
In GitLab by @jvande42b on Jul 13, 2018, 14:57

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 43](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR43)



If you run the cascading channel example with a multiplier of [0.01, 0.05, 0.1, 0.5, 1.0], you will find that as you approach 1.0 the scaling continuously progresses from hardly smoothed at all to nice and smooth, but with an increasing number of iterations to find a solution. At some point greater than 1.0, you will cross the max_iter cut-off. This is expected, because the flexibility is limited at the prio this goal runs at. I found that 0.1 was the ideal multiplier for this model, providing both fast convergence and excellent smoothing. I imagine there will be a similar multiplier for most models, but cannot guarantee that it will the same for all. It would probably require tuning.

So this parameter helps you tune the smoothing into a sweet spot that provides sufficient smoothing while not needing hundreds or thousands of iterations.

What do you think about this approach?

SGeeversAtVortech:
In GitLab by @vreeken on Jul 13, 2018, 15:31

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 43](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR43)



I'll check that myself later, but I find that behavior somewhat suspicious of no other goals are present at that priority. Maybe I'm missing something, but if all it does is to scale the objective, results shouldn't change that much.

SGeeversAtVortech:
In GitLab by @vreeken on Jul 14, 2018, 17:19

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 43](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR43)



Tested with the goal_programming example, but as I expected nothing changes except for the value of the objective. This goes squared with the scaling factor, because of the order=2.
Testing it with the cascading example, the same holds true if you only smooth one single state. So the convergence issues you are seeing are because of the interaction of various states, possibly combined with the order=2.

SGeeversAtVortech:
In GitLab by @baayen on Jul 16, 2018, 08:15

I like the idea of making the order an option and defaulting to order=1. That may also address scaling issues.

Minimizing abs(x-y) is equivalent to min z, s.t. x-y <= z and -(x-y) <= z.

SGeeversAtVortech:
In GitLab by @baayen on Jul 16, 2018, 08:16

Commented on [src/rtctools/optimization/smooth_dynamics_mixin.py line 57](https://github.com/Deltares/rtc-tools/compare/c8feae29bf1a6a2386b52a046ad76a434e17241b..b3066a57d93eaf09c998fa04899d0ad1950ba0fb#diff-9b13b6dc4fb3d7bd7c6837c72aaf6e9cR57)



I'd suggest to stick with a flat list of variables for now.  We generally try to stay clear of weights as much as possible.

SGeeversAtVortech:
In GitLab by @vreeken on Sep 3, 2018, 18:08

Another idea is to define `priority_completed` as well, and let that (dynamically) calculate the value of the nominal (such that the objective is on the order of 1). Then we avoid scaling issues where the bounds are very wide to the typically observed changes.

SGeeversAtVortech:
In GitLab by @vreeken on Sep 6, 2019, 13:14

SGeeversAtVortech:
In GitLab by @SGeeversAtVortech on Dec 8, 2022, 14:38

Closed, since this MR has been inactive for quite some time.

PR #1576, opened 2024-10-07
Optimize checks

SGeeversAtVortech:
In GitLab by @gijsber on Aug 20, 2018, 16:40

CSV Mixin: collect log info before writing log message
Goal programming Mixin: improve performance creating subproblems (one loop through goals instead of two)

Je moet Commit 8ef96431 hebben


SGeeversAtVortech:
In GitLab by @vreeken on Aug 23, 2018, 21:43

closed

PR #1582, opened 2024-10-07
WIP: History support CSVMixin

### LINKED ISSUE #1010: Let user transparently access datetime of optimization series
In GitLab by @vreeken on Feb 27, 2018, 17:16

Currently in `pre()` and `post()`, there is no supported way to ask for simulation datetimes, e.g. like `self.times()` returns seconds. Having access to datetimes could be nice for general optimation rules ('date between ..'), but even more so for things like plots or custom data output. One can use e.g. `self._CSVMixin__timeseries_times` instead (or the equivalent of `PIMixin`), but this is a very hacky (not public API) way of doing so.

It would also be nice to support (something like?) `self.get_timeseries()` that returns a timeseries with datetimes+values instead of seconds+values.

Some related questions:
1. Do we support input timeseries that do _not_ have a datetime field, but e.g. just seconds?
2. If so, what would we return if we were to ask for `self.datetimes()`. Nothing, of start from year 0?

SGeeversAtVortech:
In GitLab by @AnneHommelberg on Oct 10, 2018, 10:18

Fixes #1010.

The optimization CSVMixin now supports history, similar to the PIMixin. The forcast date can be set through the csv_forecast_date field. If no forecast date is set, the first time of the input timeseries is used (no history).

Important note: previously the default history was used, which uses the initial_state. Now, by default the first values in the input timeseries are used. Since the initial_state is read from a separate file in the case of the CSVMixin this could result in different values.
Should this be changed back to using the initial_state as a default?


SGeeversAtVortech:
In GitLab by @TPiovesan on Oct 10, 2018, 15:21

It should be change back to using initial_state as default. @vreeken, right?

SGeeversAtVortech:
In GitLab by @TPiovesan on Oct 10, 2018, 15:34

Commented on [src/rtctools/optimization/csv_mixin.py line 324](https://github.com/Deltares/rtc-tools/compare/854d9605ca0ae1e567f6c308935f83799b61c515..141ae703de6b6af936f2099469b9d97c16ba8604#diff-cf725ba9480b9b88b0a4ef19aca75b34L324)



In 
`def seed():`
you need to crop the timeseries such that it does not include the history. So

`s = Timeseries(self.__timeseries_times_sec[self.__forecast_index:]
, self.__timeseries[ensemble_member][variable][self.__forecast_index:]
)
`

This will also fix the problem in `def test_seed`.

SGeeversAtVortech:
In GitLab by @AnneHommelberg on Oct 12, 2018, 09:09

@TPiovesan Is that a bug in the PIMixin then? Because I haven't changed that code and there it is not cropped either.

SGeeversAtVortech:
In GitLab by @vreeken on Dec 16, 2018, 20:07

@AnneHommelberg Can this MR be closed, as it's superseded by !227 ?

SGeeversAtVortech:
In GitLab by @AnneHommelberg on Dec 17, 2018, 08:55

@vreeken Yes, !227 achieves the same thing.

SGeeversAtVortech:
In GitLab by @AnneHommelberg on Dec 17, 2018, 08:55

closed

PR #1597, opened 2024-10-07
WIP: GP: fix minimization goals

SGeeversAtVortech:
In GitLab by @TPiovesan on Jun 17, 2019, 12:06

* [ ]  Test if problems are adversely affected by this change

Fix how minimization goals are created when keep_soft_constraints=False. New reformulation allows for a larger feasible space and a proper application of goal relaxation.

Add unit tests to check how minimization goals constraints and the effect of goal relaxation.

### DIFF
--- a/src/rtctools/optimization/goal_programming_mixin.py
+++ b/src/rtctools/optimization/goal_programming_mixin.py
@@ -1120,13 +1120,12 @@ class GoalProgrammingMixin(OptimizationProblem, metaclass=ABCMeta):
         else:
             # Epsilon is equal to the function value
             if options['fix_minimized_values'] and goal.relaxation == 0.0:
-                m = epsilon / goal.function_nominal
-                M = epsilon / goal.function_nominal
+                m = M = epsilon**goal.order / goal.function_nominal
                 self.check_collocation_linearity = False
                 self.linear_collocation = False
             else:
                 m = -np.inf * np.ones(epsilon.shape)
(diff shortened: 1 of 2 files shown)


SGeeversAtVortech:
In GitLab by @vreeken on Jun 19, 2019, 13:56

added 2 commits

<ul><li>bcaab7b4 - GPMixin: Update comment for hard constraint of minimization goals</li><li>a2e653a2 - GPMixin: Fix hard constraint for minimization goals</li></ul>

[Compare with previous version](/deltares/rtc-tools/merge_requests/290/diffs?diff_id=45739732&start_sha=237f8f699f13c10dd141e955310995bab9496218)

SGeeversAtVortech:
In GitLab by @vreeken on Jun 19, 2019, 13:59

@baayen > I'd like to make sure that you didn't deliberately make the decision to exclude the goal order for some reason. To me it make sense (for minimization goals) to set a constraint on the exact thing we solved for, i.e. ``function**order``, and not on just ``function``.

SGeeversAtVortech:
In GitLab by @baayen on Jun 26, 2019, 13:18

The downside of using `function**order` in the constraints is that the constraints become non-linear.  I would do some testing to make sure that this change does not adversely affect performance of existing models.

SGeeversAtVortech:
In GitLab by @vreeken on Aug 9, 2019, 15:36

PR #1607, opened 2024-10-07
WIP: Fix initial_state() and history()

### LINKED ISSUE #1123: IOMixin fails to fill with NaN when calling set_timeseries
In GitLab by @vreeken on Sep 7, 2019, 13:37

Related to #1121, but now for 2.4/master.

## Calling `set_timeseries` with a `Timeseries` object
The current check is

```python
            if not np.array_equal(timeseries_times_sec, timeseries.times):
```

This should be a superset check. Note that the filling with NaNs code was/is already there, but it just couldn't get there.

## Calling `set_timeseries` with an `np.ndarray` object

```python
            if check_consistency and len(self.times()) != len(timeseries):
```
Should also check that `timeseries_times_sec` is a superset of `self.times()`

SGeeversAtVortech:
In GitLab by @vreeken on Sep 17, 2019, 22:35

* [x]  On top of !323 

* [ ]  After some more discussion, it actually seems better to just remove initial_state. There are so many things that can be very confusing when keeping both around.

Before we had the supposed default implementation in OptimizationProblem
that history() also returned values in initial_state(), and vice versa.
This would never actually work properly, as one would end up in an
endless loop.

The reason this bug was not encountered until this point, is that
ModelicaMixin did _not_ call super() for initial_state(). Furthermore,
PIMixin/IOMixin did _not_ call super() either. So in practice we never
got the behavior that history() referred to values in intial_state() and
vice versa.

The "mother" method of the two is history(), as this is the only method
actually called in CollocatedIntegratedOptimizationProblem. This means
that when a user uses any IOMixin class, but implements their own
initial_state() method, this method would never actually be called.

To fix this bug, we remove the supposed referral of initial_state() to
history() (which was not there in practice), but we keep the call to
initial_state() in OptimizationProblem's implementation of history(). We
also make sure that all classes properly call their respective super()
methods.

Closes #1123

### DIFF
--- a/src/rtctools/optimization/io_mixin.py
+++ b/src/rtctools/optimization/io_mixin.py
@@ -7,7 +7,6 @@ import casadi as ca
 
 import numpy as np
 
-from rtctools._internal.alias_tools import AliasDict
 from rtctools._internal.caching import cached
 from rtctools.optimization.optimization_problem import OptimizationProblem
 from rtctools.optimization.timeseries import Timeseries
@@ -219,7 +218,7 @@ class IOMixin(OptimizationProblem, metaclass=ABCMeta):
     @cached
     def history(self, ensemble_member):
         # Load history
-        history = AliasDict(self.alias_relation)
+        history = super().history(ensemble_member)
 
         end_index = bisect.bisect_left(self.io.times_sec, self.initial_time) + 1
 
--- a/src/rtctools/optimization/modelica_mixin.py
+++ b/src/rtctools/optimization/modelica_mixin.py
@@ -207,7 +207,7 @@ class ModelicaMixin(OptimizationProblem):
 
     @cached
     def initial_state(self, ensemble_member):
-        initial_state = AliasDict(self.alias_relation)
+        initial_state = super().initial_state(ensemble_member)
 
         # Parameter values
         parameters = self.parameters(ensemble_member)
--- a/src/rtctools/optimization/optimization_problem.py
+++ b/src/rtctools/optimization/optimization_problem.py
@@ -399,7 +399,7 @@ class OptimizationProblem(DataStoreAccessor, metaclass=ABCMeta):
 
     def history(self, ensemble_member: int) -> AliasDict[str, Timeseries]:
         """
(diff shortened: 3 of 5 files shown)


SGeeversAtVortech:
In GitLab by @vreeken on Sep 17, 2019, 22:55

added 1 commit

<ul><li>6afbdd8a - Fix initial_state() and history()</li></ul>

[Compare with previous version](/deltares/rtc-tools/merge_requests/322/diffs?diff_id=55411741&start_sha=c8edde990cd9969ad4320884d4c224b6745d22fb)

SGeeversAtVortech:
In GitLab by @vreeken on Sep 18, 2019, 24:03

SGeeversAtVortech:
In GitLab by @vreeken on Sep 26, 2019, 17:19

unmarked as a **Work In Progress**

SGeeversAtVortech:
In GitLab by @vreeken on Sep 26, 2019, 17:19

added 9 commits

<ul><li>6afbdd8a...ba9bdddb - 4 commits from branch <code>master</code></li><li>fe9afdf2 - Simulation/CSVMixin: Fixes related to IO refactor</li><li>b79ad555 - Simulation: Fix reference to results storage</li><li>b6ebed90 - Tests: Fix NetCDFMixin write test</li><li>b8f4ff8a - Add test for the examples; CI runs for tags</li><li>f3e31900 - Fix initial_state() and history()</li></ul>

[Compare with previous version](/deltares/rtc-tools/merge_requests/322/diffs?diff_id=56532711&start_sha=6afbdd8a2477dcaf292f455ded2b8a364e105aba)

SGeeversAtVortech:
In GitLab by @vreeken on Sep 26, 2019, 17:29

marked the checklist item **On top of !323** as completed

SGeeversAtVortech:
In GitLab by @vreeken on Oct 9, 2019, 21:28

SGeeversAtVortech:
In GitLab by @vreeken on Oct 17, 2019, 15:14

Closed in favor of !343

SGeeversAtVortech:
In GitLab by @vreeken on Oct 17, 2019, 15:14

closed

PR #1638, opened 2024-10-08
Seed previous run

SGeeversAtVortech:


src/rtctools/optimization/csv_mixin.py
```diff
@@ -219,6 +219,85 @@ def check_initial_state_array(initial_state):
                             "Set csv_equidistant = False if this is intended.".format(times[i + 1])
                         )
 
+    def seed_with_imported_result(self, seed, current_ensemble_member):
```
SGeeversAtVortech, line 222: This seems like quite a complicated file for reading a seed from a csv file and I suspect it has some code duplication. Aren't the functionalities for reading a csv file not already there?

src/rtctools/optimization/pi_mixin.py
```diff
@@ -153,6 +156,90 @@ def read(self):
                         check_duplicates=self.pi_check_for_duplicate_parameters,
                     )
 
+    def seed_with_imported_result(self, seed, current_ensemble_member):
```
SGeeversAtVortech, line 159: I have the same concern here as for the csv_,mixin version.

src/rtctools/optimization/homotopy_mixin.py
```diff
@@ -137,24 +137,35 @@ def optimize(self, preprocessing=True, postprocessing=True, log_solver_failure_a
...
-                        "be decreased further, as that would violate the minimum delta "
-                        "theta of {}.".format(self.__theta, options["delta_theta_min"])
-                    )
-                    if log_solver_failure_as_error:
-                        logger.error(failure_message)
+                    # if an imported seed was used and theta != 0.0 then we fallback to using
+                    # default homotopy approach
+                    if self.seeding_options()["import_seed"] and 0.0 < options["theta_start"]:
+                        self.seeding_failed = True
+                        self.__theta = -delta_theta
```
SGeeversAtVortech, line 144: Do you want to decrease self.__theta if it already equals theta_start? I would expect you just keep it unchanged in this particular case.

src/rtctools/optimization/homotopy_mixin.py
```diff
@@ -137,24 +137,35 @@ def optimize(self, preprocessing=True, postprocessing=True, log_solver_failure_a
...
-                    failure_message = (
-                        "Solver failed with homotopy parameter theta = {}. Theta cannot "
-                        "be decreased further, as that would violate the minimum delta "
-                        "theta of {}.".format(self.__theta, options["delta_theta_min"])
-                    )
-                    if log_solver_failure_as_error:
-                        logger.error(failure_message)
+                    # if an imported seed was used and theta != 0.0 then we fallback to using
+                    # default homotopy approach
+                    if self.seeding_options()["import_seed"] and 0.0 < options["theta_start"]:
```
SGeeversAtVortech, line 142: Shouldn't this if statement also include something like `and not self.seeding_failed`?

SGeeversAtVortech (commented):
I didn't really look into the fallback option in goal_programming. I have some concerns with the seed functions in csv_mixin and pi_mixin. I think we can reuse more existing functionalities. Also, it seems in many cases that we assume that the seed is the result from a previous run, while we could keep this more general and let it just be a seed from a separate file.

SGeeversAtVortech (commented):
A few questions/remarks:

- Should we have a separate option for the fallback? Maybe sometimes you don't want to fallback to a case without seed,
- In case of homotopy, I think it is clearer that we provide separate homotopy parameters if we use a different theta_start for example.

Ailbhemit:
approach with seed_mixin is used instead

PR #1640, opened 2024-10-15
workflows: add workflow for publishing version on pypi

SGeeversAtVortech:
Add a workflow to publish an rtc-tools version on pypi when pushing a tag starting with "2.".


SGeeversAtVortech:
The issue has been resolved in #1650.

PR #1642, opened 2024-11-12
Add a SeedMixin class for seed options

SGeeversAtVortech:
Add a SeedMixin class for enabling seed options such as reading a seed from a file and falling back to a default seed if the solver failed.


src/rtctools/data/timeseries.py
```diff
@@ -0,0 +1,86 @@
...
+    :param values:  1D array of values with the same length as times.
+    :interp_args:   Dict of arguments passed to numpy.interp.
+
+    :returns:       List of values where nans have been replaced with interpolated values.
+    """
+    if interp_args is None:
+        interp_args = {}
+    times_sec = np.array([(t - times[0]).total_seconds() for t in times])
+    nans = np.isnan(values)
+    result = np.interp(times_sec, times_sec[~nans], values[~nans], **interp_args)
```
Ailbhemit, line 85: This leads to an error if the seed is filled with nans. Instead we can already return the function earlier if this is the case 
if nans.all():
        return values  # Return early if all values are NaN
return np.interp(times_sec, times_sec[~nans], values[~nans], **interp_args)

Ailbhemit (changes requested):
Hi Sjoerd. This works well, meets the requirements and is user friendly. There was one bug I found during testing where the seed cannot be interpolated due to being filled with nans.

Ailbhemit:
I am also not able to see how a user can choose whether or not to use a seed based upon and input parameter? Parameters are not yet available when seed_options is accessed.

SGeeversAtVortech:
> I am also not able to see how a user can choose whether or not to use a seed based upon and input parameter? Parameters are not yet available when seed_options is accessed.

The user can overwrite the seed_options method, similar to how we handle other options. By default, option seed_timeseries is None, which means no timeseries is used for the seed.

src/rtctools/data/timeseries.py
SGeeversAtVortech: Yep, I'll fix it and include this case in the unit test.

Ailbhemit:
> > I am also not able to see how a user can choose whether or not to use a seed based upon and input parameter? Parameters are not yet available when seed_options is accessed.
> 
> The user can overwrite the seed_options method, similar to how we handle other options. By default, option seed_timeseries is None, which means no timeseries is used for the seed.

This is now possible (accessing parameters) with the recent commit

Ailbhemit (approved):
PR is now ready to merge

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+
+    seed_timeseries: Union[Path, None] = None
+    is_binary: bool = False
+    validate_timeseries: bool = False
+    times_are_equidistant: bool = False
+    fallback: bool = False
+
+
+class SeedMixin(OptimizationProblem):
```
jbaayen, line 50: I'm still not clear on the purpose of this new mixin.

Currently, if a time series is specified for an optimization variable in `timeseries_import.csv`, its values are taken as seed values.

All this mixin seems to do is provide a new pathway to specify seed values on top and above of the existing pathway. It would therefore seem to contradict the "Zen of Python" which states that "There should be one-- and preferably only one --obvious way to do it." (https://peps.python.org/pep-0020/).

tests/optimization/data/reservoir/reservoir.mo
```diff
@@ -0,0 +1,9 @@
+model Reservoir
+    // Basic model for in/outlow of a reservoir
+    parameter Real theta;
+    input Real Q_in(fixed=true);
+    input Real Q_out(fixed=false, min=0.0, max=5.0);
+    output Real Volume;
```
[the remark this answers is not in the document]
SGeeversAtVortech: I'll switch to lower case. Capital letters are used in many of the other examples in rtc-tools though.

src/rtctools/data/timeseries.py
```diff
@@ -0,0 +1,88 @@
+"""Module for processing timeseries data."""
```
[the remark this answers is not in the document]
SGeeversAtVortech: There is a Timeseries class in `rtctools.optimization`, but the new functionalities in this PR are not specific for optimization but could also be used for simulation. Therefore, I figured it makes more sense to put them in `rtctools.data`. Perhaps it would also make sense to move the Timeseries class to `rtctools.data`, but that would be a breaking change.

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,126 @@
+"""Module for seeding functionalities.
```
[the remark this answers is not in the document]
SGeeversAtVortech: I'll move the comments.

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,126 @@
...
+    fallback: bool = False
+
+
+class SeedMixin(OptimizationProblem):
+    """
+    Adds options for reading a seed from a given file to your optimization problem.
+    """
+
+    def __init__(self, **kwargs):
+        self._seed_timeseries: Path = None
```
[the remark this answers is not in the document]
SGeeversAtVortech: I think it is, since `_seed_timeseries` is an attribute of `SeedMixin`. It indicates which file, if any, is used for the seed.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: > I'm still not clear on the purpose of this new mixin.
> 
> Currently, if a time series is specified for an optimization variable in `timeseries_import.csv`, its values are taken as seed values.
> 
> All this mixin seems to do is provide a new pathway to specify seed values on top and above of the existing pathway. It would therefore seem to contradict the "Zen of Python" which states that "There should be one-- and preferably only one --obvious way to do it." (https://peps.python.org/pep-0020/).

From what I understand, timeseries values for optimization variables are currently converted into constraints, while values for states are converted into seeds. We have cases where we would like to mix these two, for example to have a constraint for the first time step of a state and a seed for all the other timesteps, or to use constraints on a control variable for certain time steps and a seed for others. This seems currently not possible.

It is also not obvious currently if a timeseries value will be used as a seed or a constraint. Using a separate seed file makes it more explicit what will be used for a seed. It also allows to easily switch between seeds without having to copy all other timeseries. The seed values are also different from constraints / fixed input values, since they do not define the problem, but are more a solver option.

We agree that ideally there should be one obvious way to do it. Perhaps we can decide that for the future, the seed should always be defined in a separate file, but to avoid breaking changes, we still allow seeds in the timeseries_import file.

SGeeversAtVortech: @jbaayen: before I address the other comments, I think it's convenient if we first resolve this issue. Please let me know if you agree with the points above.

jbaayen: @SGeeversAtVortech what makes you think that provided timeseries values for optimization variables are turned into constraints?

jbaayen: AFAIK the only case where a time series is used as a constraint is when it is labeled 'variable_name + "_Min"' or 'variable_name + "_Max"', in which case the values are used to construct lower respectively upper bounds for the named variable.

Ailbhemit: It is also added as a constraint if the variable is a controlled input

SGeeversAtVortech: I did some code digging and I think the confusion comes from src\rtctools\optimization\collocated_integrated_optimization_problem.py in the part that starts with `# Initial conditions specified in history timeseries`:

```python
            # Initial conditions specified in history timeseries
            history = self.history(ensemble_member)
            for variable in itertools.chain(
                self.differentiated_states, self.algebraic_states, self.controls
            ):
                try:
                    history_timeseries = history[variable]
                except KeyError:
                    pass
                else:
                    interpolation_method = self.interpolation_method(variable)
                    val = self.interpolate(
                        t0,
                        history_timeseries.times,
                        history_timeseries.values,
                        np.nan,
                        np.nan,
                        interpolation_method,
                    )
                    val /= self.variable_nominal(variable)

                    if not np.isnan(val):
                        idx = self.__indices_as_lists[ensemble_member][variable][0]

                        if val < lbx[idx] or val > ubx[idx]:
                            logger.warning(
                                "Initial value {} for variable '{}' outside bounds.".format(
                                    val, variable
                                )
                            )

                        lbx[idx] = ubx[idx] = val
```

It seems that values from before and including the initial time are converted into constraints and later values are used as a seed.

SGeeversAtVortech: In any case, the feature that we try to implement here is to select a seed by filename and automatically try another seed when the optimization fails. Currently, the user would still need to carefully copy+paste historic data of the seeded timeseries + all data of the other timeseries to a separate file and rerun the optimization from scratch, which is rather cumbersome and prone to errors. So although it is technically possible to switch seed, I would not say there is an obvious way to do this right now.

jbaayen: This is almost correct: the value from the time series at t0 will be used as initial condition. But there are no constraints created for t < t0 and for t > t0.

The appropriate way to switch seed during a run is to override the `seed()` method. An example of a `seed()` override implementation can be found in `GoalProgrammingMixin`.

SGeeversAtVortech: The user could of course implement his own `seed()` and `optimize()` function as done here in `SeedMixin`, but this implementation is not trivial and has a few pitfalls. The point of this feature, and most other features in rtc-tools, is to make life easier for the user. Probably you can solve any optimization problem with just the `OptimizationProblem` class, but the other mixins are there to provide useful tools and reduce the amount of work on the users side.

jbaayen: Yes, but we don't provide dedicated mixins to override each and every overridable function individually. If we did that, we'd have 50 or so different mixins.

The design philosophy has been that mixins add new significant and broadly reusable pieces of functionality that are not available in the core classes.

May I ask you to again try to explain in a few paragraphs what exactly the value-add of the SeedMixin is, considering explicitly that it is currently already possible to provide seed data in a user-friendly manner (e.g., via timeseries_import.csv)?

SGeeversAtVortech: The way I see it, the SeedMixin takes care of running/rerunning an optimization problem for different seeds. It enables the following.

- Enable to select and switch between different seeds, without having to redefine the problem input (I would not consider seeds problem input, since they do not affect the problem definition).
- Enable automatically rerunning if a seed failed and enable adding logic to select the next seed.

For the first property, we would now need to copy all other timeseries in timeseries_import.csv and be careful with the initial values just to use another seed. For the second property, the user would now need to manually check why a run failed (because of the seed or not) and then rerun for different input if necessary.

I agree that we should not add a new mixin when the same can be achieved with a few single functions and the current mixins.

jbaayen: I think you are saying that the core value-add is the support of multiple seeds per run, which introduces robustness for cases where the first seed does not result in a convergent run. This makes sense.

How about renaming the mixin to 'MultiSeedMixin' to stress this point?

src/rtctools/data/timeseries.py
```diff
@@ -0,0 +1,88 @@
...
+import numpy as np
+
+import rtctools.data.csv as csv
+import rtctools.data.pi as pi
+import rtctools.data.rtc as rtc
+
+
+def get_timeseries_from_csv(
+    file: Path, delimiter=","
+) -> Tuple[List[datetime], Dict[str, np.ndarray]]:
```
jbaayen, line 15: This method is specific to CSV files and so should be in 'CSVMixin'

src/rtctools/data/timeseries.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    data = csv.load(file, delimiter=delimiter, with_time=True)
+    times = data[data.dtype.names[0]]
+    values = {key: np.asarray(data[key], dtype=np.float64) for key in data.dtype.names[1:]}
+    return times, values
+
+
+def get_timeseries_from_pi(
+    file: Path, data_config=rtc.DataConfig, is_binary=False, validate=False
+) -> Tuple[List[datetime], Dict[str, np.ndarray]]:
```
jbaayen, line 31: This method is specific to PI files and so should be in 'PIMixin'

src/rtctools/data/timeseries.py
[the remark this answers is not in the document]
jbaayen: If 'get_timeseries_from_{csv,pi}' move back to the format-specific modules, the remaining module could be renamed to 'util.py'. This would be less confusing than having multiple modules named 'timeseries' within the same package.

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+    def seed_timeseries(self) -> Path:
+        """Get the path of the seed timeseries."""
+        return self._seed_timeseries
+
+    def seed(self, ensemble_member):
+        if not self._use_seed_timeseries():
+            return super().seed(ensemble_member)
+        seed: dict = super().seed(ensemble_member).copy()  # Copy to prevent updating cached seeds.
+        if isinstance(self, CSVMixin):
+            times, values_dict = get_timeseries_from_csv(self._seed_timeseries)
```
jbaayen, line 90: This code introduces a dependency between 'SeedMixin' and 'CSVMixin', which implicitly assumes that the 'SeedMixin' would only be used in combination with either 'CSVMixin' or 'PIMixin'. As a result, it would not work well with other, 3rd party IO mixins.

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        elif isinstance(self, PIMixin):
```
jbaayen, line 91: This code introduces a dependency between 'SeedMixin' and 'PIMixin', which implicitly assumes that the 'SeedMixin' would only be used in combination with either 'CSVMixin' or 'PIMixin'. As a result, it would not work well with other, 3rd party IO mixins.

src/rtctools/optimization/seed_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        super().__init__(**kwargs)
+
+    def _use_seed_timeseries(self):
+        """Return True if a seed from a timeseries is used for the current run."""
+        if self._seed_timeseries is None:
+            return False
+        if isinstance(self, GoalProgrammingMixin):
+            if not self._gp_first_run:
+                return False
+        if isinstance(self, HomotopyMixin):
```
jbaayen, line 34: Ideally there would be no horizontal dependencies between mixins.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: A 'MultiSeedMixin' sounds reasonable.

src/rtctools/data/timeseries.py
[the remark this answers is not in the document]
SGeeversAtVortech: Sounds reasonable. I'll rename this file and move the read functions to their respective Mixin module.

src/rtctools/data/timeseries.py
SGeeversAtVortech: I'll move it.

src/rtctools/data/timeseries.py
SGeeversAtVortech: I'll move it.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: Yes, it doesn't extend nicely if a user would use another file type. I'll try to find another approach.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: Yes true, although I could not find an easy solution for this. We use a similar approach in HomotopyMixin although it is not ideal there either. I'll try again if I can get rid of the dependencies.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: The user is now responsible for implementing the method seed_from_file. Using the get_timeseries functions, this requires only a few lines.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: The user can now overwrite the `use_seed_file` method (by default returns True). This avoids mixin dependencies and only requires a few lines of code from the user.

src/rtctools/optimization/seed_mixin.py
SGeeversAtVortech: The user can now overwrite the use_seed_file method (by default returns True). This avoids mixin dependencies and only requires a few lines of code from the user.

doc/optimization/seeding.rst
```diff
@@ -0,0 +1,6 @@
+Seeding
+=======
+
+.. autoclass:: rtctools.optimization.seed_mixin.SeedMixin
```
jbaayen, line 4: Refer to MultiSeedMixin.

It would also be good to discuss here that standard, "single" seeding functionality is available without requiring this mixin.

src/rtctools/optimization/pi_mixin.py
```diff
@@ -290,3 +294,28 @@ def set_unit(self, variable: str, unit: str):
         ), "set_unit can only be called after read() in pre() has finished."
         self.__timeseries_import.set_unit(variable, unit, 0)
         self.__timeseries_export.set_unit(variable, unit, 0)
+
+
+def get_timeseries_from_pi(
+    file: Path, data_config=rtc.DataConfig, is_binary=False, validate=False
+) -> Tuple[List[datetime], Dict[str, np.ndarray]]:
```
jbaayen, line 301: If this method is only used for testing, it should not be included here

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+        """Return True if the seed file is used for the current run.
+
+        By default, this is True.
+        Overwrite this method when in some cases the seed is not based on the seed file,
+        e.g. in case of goal programming, return False after the first priority,
+        or in case of homotopy, return False after the homotopy parameter is increased.
+        """
+        return True
+
+    def seed_file(self) -> Optional[Path]:
```
jbaayen, line 27: What is this method used for?

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+        """Get the selected seed file.
+
+        If the selected seed file is None, use the default seed.
+        """
+        return self.__seed_file
+
+    def seed_files(self) -> Iterable[Optional[Path]]:
```
jbaayen, line 34: Should probably be a list of identifiers rather than files, as the seeds do not necessarily need to come from files stored on disk.

jbaayen (changes requested):

doc/optimization/seeding.rst
SGeeversAtVortech: Good catch! I'll update and double check the documentation

src/rtctools/optimization/pi_mixin.py
SGeeversAtVortech: This method seems to have little added value over the current pi.Timeseries class, so I'll just remove it.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: I have renamed the methods somewhat. This will now return the selected seed id (before the selected seed file).

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: Yes, I'll make it a bit more generic and use seed identifiers instead.

doc/optimization/multi_seeding.rst
```diff
@@ -0,0 +1,10 @@
+Using multiple seeds
+====================
+
+A seed can be defined by implementing the :py:meth:`OptimizaionProblem.seed` method.
```
jbaayen, line 4: There's a typo in `OptimizaionProbelm.seed`

src/rtctools/data/util.py
```diff
@@ -0,0 +1,42 @@
+"""Module with utilities for processing data."""
```
jbaayen, line 1: We typically don't have this first-line comments in RTC-Tools modules

src/rtctools/optimization/pi_mixin.py
```diff
@@ -278,6 +271,13 @@ def timeseries_export(self):
         """
         return self.__timeseries_export
 
+    @property
+    def data_config(self):
```
jbaayen, line 275: This addition should not be part of this PR

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        raise NotImplementedError("This method should be implemented by the user.")
+
+    def seed_from_id(self, seed_id) -> Dict[str, Timeseries]:
+        """
+        Get the seed timeseries from the selected identifier.
+
+        This method should be implemented by the user.
+        """
+        del seed_id
```
jbaayen, line 49: What is this line needed for?

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,79 @@
...
+        Return ``True`` if the seed corresponding to the selected seed identifier is used
+        for the current run.
+        By default, this is ``True``.
+        Overwrite this method when in some cases the seed is not based on the seed id,
+        e.g. in case of goal programming, return False after the first priority,
+        or in case of homotopy, return ``False`` after the homotopy parameter is increased.
+        """
+        return True
+
+    def selected_seed_id(self) -> Any:
```
jbaayen, line 28: Do we need this public method? Isn't it sufficient to maintain the private `__selected_seed_id` variable?

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,79 @@
...
+class MultiSeedMixin(OptimizationProblem):
+    """
+    Enables a workflow to solve an optimization problem by trying multiple seeds.
+    """
+
+    def __init__(self, **kwargs):
+        self.__selected_seed_id = None
+        super().__init__(**kwargs)
+
+    def use_seed_id(self) -> bool:
```
jbaayen, line 16: Is this method needed?

For goal programming, couldn't the logic be simplified to:
- if preprocessing is True, select the seed from the seed IDs
- if preprocessing is False, use the seed() implementation from the superclass

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,79 @@
...
+        return True
+
+    def selected_seed_id(self) -> Any:
+        """Get the selected seed identifier.
+
+        If the selected seed id is None, use the default seed.
+        """
+        return self.__selected_seed_id
+
+    def seed_ids(self) -> Iterable:
```
jbaayen, line 35: A simpler way to implement these two methods would be to replace them with a single `seeds` method that return a list of seed dictionaries.

jbaayen (changes requested):

doc/optimization/multi_seeding.rst
SGeeversAtVortech: Yep, will fix this.

src/rtctools/data/util.py
SGeeversAtVortech: I'll remove it.

src/rtctools/optimization/pi_mixin.py
SGeeversAtVortech: We do need this when reading a seed from a given xml file. It is also used in the MultiSeedMixin test. I will move it to a separate commit and perhaps make a separate PR out of it.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: To indicate that this parameter is not used in this function.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: Perhaps not. I'll see if this method can be avoided.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: I'm not sure if I understand what you mean. When calling `super().optimize` from `MultiSeedMixin`, the argument `preprocessing` will always be set to `False`. We somehow need to determine if we want to use an initial seed (e.g. when solving for the first priority) or a seed based on the last results (e.g. when solving for the next priorities).

src/rtctools/optimization/multi_seed_mixin.py
jbaayen: Do we use this pattern elsewhere in the code too? Or is it just here?

SGeeversAtVortech: It seems to appear only here. I can also remove it.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: In some cases, we might want to use different solver parameters depending on the seed. For example, when using homotopy, we might want to start immediately with the nonlinear case (theta_start = 1), while for the default seed, we might want to start with the linear case (theta_start = 0). Exposing the selected seed id allows the user to adapt their solver options based on the seed. This behaviour is also checked in the SeedMixin test.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: Using separate seed ids allows the user to select specific solver options based on the seed id as also mentioned in the `selected_seed_id` discussion. It therefore seems to have added value to keep `seed_ids` and `seed_from_id` separated.

src/rtctools/optimization/pi_mixin.py
SGeeversAtVortech: I've moved it to a separate commit.

src/rtctools/optimization/multi_seed_mixin.py
jbaayen: why not override solver_options() and return the desired solver depending on the current value of theta? If it is already possible to implement the desired behavior in a lean way, we should not implement a new way to do the same.

src/rtctools/optimization/multi_seed_mixin.py
jbaayen: See my comments above - I think there is an existing way to achieve the desired solver options.

src/rtctools/optimization/multi_seed_mixin.py
jbaayen: By default, optimize() is called with preprocessing set to True. It's only in GoalProgrammingMixin that the method is overridden with an implementation that calls optimize() multiple times with preprocessing=False. You can use this to override the top-level implementation with a preprocessing=True/False check.

SGeeversAtVortech: The way `MultiSeedMixin.optimize()` is now implemented is that it will call pre at the start if preprocessing is True and within a loop for each seed call `super().optimize` with preprocessing is False. If MultiSeedMixin is the top level, it has no idea if GoalProgrammingMixin is at its first run or not, so we leave this logic up to the user by allowing him to override this method.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: The case we had in mind, and which is included the test, has theta_start depend on the seed, not the other way around. We solve this by overwriting the homotopy options and let theta_start depend on the seed_id.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: I have to admit that during development, I had in mind that `MultiSeedMixin` would be the top-level class in the inheritance hierarchy although we never impose this. Perhaps by making it sub-level to HomotopyMixin/GoalProgrammingMixin the logic can indeed be simplified. Let me check this.

SGeeversAtVortech:
@jbaayen : I created an alternative implementation that simplifies MultiSeedMixin and assumes it is at a lower lever in the hierarchy than GoalProgrammingMixin/HomotopyMixin. This does mean some extra work on the user side if the homotopy parameter theta_start depends on the seed, so I'm still not sure if this is the optimal solution.

src/rtctools/optimization/multi_seed_mixin.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, **kwargs):
+        self.__selected_seed = None
+        super().__init__(**kwargs)
+
+    def seeds(self) -> Iterable[Optional[Dict[str, Timeseries]]]:
+        """Return a list or iterator of seeds.
+
+        When the seed is None, the default seed will be used.
+        This method should be implemented by the user.
+        """
```
jbaayen, line 21: Yes, this is much cleaner! Thank you for taking the time to refactor this.

Can we 'raise NotImplementedError' here by default, rather than returning the somewhat artificial '[None]' list?
If this method is not overridden/implemented, the user should not be using the mixin.

src/rtctools/optimization/csv_mixin.py
```diff
@@ -295,3 +279,19 @@ def write_output(ensemble_member, folder):
...
+
+    :param file:        Path to CSV.
+    :param delimiter:   CSV column delimiter.
+
+    :returns: A list of time stamps (datetimes) and dict of values.
+    """
+    data = csv.load(file, delimiter=delimiter, with_time=True)
+    times = data[data.dtype.names[0]]
+    var_dict = {key: np.asarray(data[key], dtype=np.float64) for key in data.dtype.names[1:]}
+    return times, var_dict
```
jbaayen, line 297: The refactored 'MultiSeedMixin' does not require any of the changes to 'PIMixin' and 'CSVMixin'. Should the latter changes be removed or moved to a separate PR?

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: I agree this is a lot leaner, but I feel that now the user has to do too much himself if a parameter like theta_start depends on the seed. On the other hand, having these solver parameters depend on the seed also doesn't sound very logical. I think I'll try one more thing to see if we can use multiple seeds with different solver parameters in a clean way.

src/rtctools/optimization/csv_mixin.py
SGeeversAtVortech: Yes, I can make a separate PR for that.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: I've created an alternative that fits the workflow that we had in mind a bit better: #1645.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: In the latest commit (the alternative implementation) we avoid this public method.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: In the latest commit (the alternative implementation) we avoid this public method.

src/rtctools/optimization/multi_seed_mixin.py
SGeeversAtVortech: In the latest commit (the alternative implementation) we indeed simply use the `seeds` method. This does require quite a bit more work by the user though. Instead of a MultiSeedMixin, I created a MultiRunMixin in a separate branch #1645, which can achieve the same and seems a bit more generic and user-friendly.

PR #1645, opened 2024-12-17
MultiRunMixin

SGeeversAtVortech:
Add a mixin class to enable a workflow for solving an optimization problem using multiple attempts.

- [ ] First, #1648 and #1649 need to be merged.


SGeeversAtVortech:
@jbaayen: the MultiRunMixin in this branch is an alternative to the MultiSeedMixin class and makes changing parameters like theta_start for different solve attempts look more natural. I think it slightly better fits the workflow that we were aiming for.

src/rtctools/optimization/multi_run_mixin.py
```diff
@@ -0,0 +1,45 @@
+from typing import Iterable
+
+from rtctools.optimization.optimization_problem import OptimizationProblem
+
+
+class MultiRunMixin(OptimizationProblem):
+    """
+    Enables a workflow to solve an optimization problem with multiple attempts.
+    """
```
jbaayen, line 9: This mixin calls `optimize()` in a loop, cycling through a list of user-specified run IDs. These run IDs have no other meaning. Isn't this functionality a bit thin to motivate a mixin?

SGeeversAtVortech: This is a rather simple mixin, but optimizing using multiple attempts seems useful for a wide variety of applications.

jbaayen: Definitely, but this is already possible using a vanilla Python for loop isn't it?

jbaayen: But it's already possible to do this by setting `preprocessing=postprocessing=False`, as is done in this mixin? 

Let's discuss this mixin with the broader group on the next technical coordination call.

SGeeversAtVortech: I will try to turn this into an example rather than its own mixin.

SGeeversAtVortech:
Instead of this Mixin, we use an example instead to illustrate how to create a workflow for multiple solver attempts: https://github.com/Deltares/rtc-tools/pull/1680.

PR #1647, opened 2025-01-14
simulation: fix initialization input variables

SGeeversAtVortech:
Only set input variables when calling IOMixin.__set_input_variables.

### DIFF
--- a/src/rtctools/simulation/io_mixin.py
+++ b/src/rtctools/simulation/io_mixin.py
@@ -94,7 +94,7 @@ class IOMixin(SimulationProblem, metaclass=ABCMeta):
             self.__cache_loop_timeseries = {}
 
             timeseries_names = set(self.io.get_timeseries_names(0))
-            for v in self.get_variables():
+            for v in self.get_input_variables():
                 if v in timeseries_names:
                     _, values = self.io.get_timeseries_sec(v)
                     self.__cache_loop_timeseries[v] = values


SGeeversAtVortech:
Currently, `IOMixin.__set_input_variables` will loop over all variables with an import timeseries. This can include an output variable for which an initial value is provided.

A problem that we came across is that  `IOMixin.__set_input_variables` will call `OptimizationProblem.set_var(variable, value)` and this `set_var` will divide `value` by a nominal and then use the updated value in `OptimizationProblem.__state_vector[index] = value`. For an output variable, the nominal might be a casadi MX symbol, and then this last statement will raise a RunTimeError of the form "'to_double' not defined for class ... ".

SGeeversAtVortech:
@jackvreeken could you have a look at this single-line PR? It is supposed to be a bugfix, but perhaps it is also a breaking change?

k-horvath-deltares (approved):

SGeeversAtVortech:
@jackvreeken could you have a look at this single-line PR? It is supposed to be a bugfix, but perhaps it is also a breaking change?

jackvreeken:
I haven't thought about the consequences that much, but if the cause is "nominal is MX", shouldn't _that_ be fixed?

SGeeversAtVortech:
> I haven't thought about the consequences that much, but if the cause is "nominal is MX", shouldn't _that_ be fixed?

The symbolic nominal seems to be a feature rather than a bug. The nominal can be expressed in terms of parameters and are evaluated in `SimulationProblem.initialize`, which is called after `IOMixin,__set_input_variables`.

It also seems natural to only set input variables in a function called `__set_input_variables`, but I'm not sure if this can cause serious breaking changes.

SGeeversAtVortech:
It is in the timeseries_import.xml since it has an initial value.

SGeeversAtVortech:
The initial value definitely should be set. In `SimulationProblem.initialize`, after evaluating the nominals, we collect initial values and call `SimulationProblem.initial_state` which in turn calls `IOMixin.timeseries_at`. I assume this is how the initial values are set.

jackvreeken:
Ah yes, so for `fixed=false` (default, which you have) it will use `initial_state` -> `self.timeseries_at` . And for `fixed=true` it will use IOMixin. Having both overlap is indeed a bit weird. 

If you can add that to the explanation in the commit message, that'd be great (now it's a bit brief). And then you can go ahead and merge

SGeeversAtVortech:
Ok, I've updated the docstring.

jarsarasty:
It can be merged now!

k-horvath-deltares (approved):

PR #1648, opened 2025-01-21
optimization: expose PI data configuration

SGeeversAtVortech:
Expose the Delft-FEWS Published Interface data configuration used for reading XML files.
This enables users to read other XML files,
e.g. for reading alternative seeds.


jackvreeken:
If I understand correctly, this means we're pinning ourselves on the "API" to FEWS-PI identifiers being `<location_id>:<parameter_id>:<qualifier_id_1>:<qualifier_id_2>:...` . Was that already the case? I don't particularly like that magic stringifying of an object/dataclass, but I also don't foresee any problems with pinning it (seeing how stable FEWS-PI is).

SGeeversAtVortech:
> If I understand correctly, this means we're pinning ourselves on the "API" to FEWS-PI identifiers being `<location_id>:<parameter_id>:<qualifier_id_1>:<qualifier_id_2>:...` . Was that already the case? I don't particularly like that magic stringifying of an object/dataclass, but I also don't foresee any problems with pinning it (seeing how stable FEWS-PI is).
> 
> Note: at some point in time I tried to make a "fews-io" Python library where I tried to abstract away this internal magic string format iirc.

I think I'm not quite following. This commit allows one to read the `DataConfig` object. It doesn't change the interface of the `DataConfig` class itself or how it reads/writes xml files.

jackvreeken:
Yes, that's exactly what I mean. You're exposing something now that used to be hidden. That means that things that _used_ to be internal-only, are now public facing. And because they're public facing, they have to remain stable.

SGeeversAtVortech:
Ok, that does not sound ideal, but isn't the xml format already pinned because of how `PIMixin` and `DataConfig` currently work? In other words, if I would change the format of my rtcDataConfig.xml file, the `PIMixin` would fail, right?

In the end, we would just like to be able to read a certain xml file (not the standard ones that are read by PIMixin). We can use pi.Timeseries for this, but this requires a DataConfig object, and for this it is useful if we can get the DataConfig object of our optimization problem.

jackvreeken:
Yes, the PIMixin would fail if you do that.

I personally feel like reading and exposing arbitrary PI-XML files is out of the scope of RTC-Tools. I see that [the fews-io package](https://pypi.org/project/fews-io/0.0.4/#history ) has recent releases; I don't know where the repository is though (can't find it on GitLab or GitHub). If that library is not sufficient for you for some reason, I'm OK with exposing DataConfig with this PR. Mostly because the internal representation has not been changed in a loooong time. 

The intent behind my original comment was to ask/clarify whether you were aware of the consequences with regards to API stability (i.e. exposing internal representations, and as a result stabilizing them).

SGeeversAtVortech:
We will investigate if we can handle xml files using the `fews-io` library.

SGeeversAtVortech:
This was required for https://github.com/Deltares/rtc-tools/pull/1645, but we now use https://github.com/Deltares/rtc-tools/pull/1680 instead.

PR #1649, opened 2025-01-21
Refactor timeseries utilities

SGeeversAtVortech:
Refactor timeseries utilities used in optimization and move some of the utilities to a more generic module.

- [ ] First, #1648 needs to be merged.


src/rtctools/data/util.py
```diff
@@ -0,0 +1,41 @@
...
+
+def check_times_are_increasing(times: List[datetime]):
+    """Check that time stamps are increasing."""
+    for i in range(len(times) - 1):
+        if times[i] >= times[i + 1]:
+            raise ValueError("Time stamps must be strictly increasing.")
+
+
+def check_times_are_equidistant(times: List[datetime]):
+    """Check that times are eqeuidistant."""
```
jackvreeken, line 15: Typo "equidistant"

src/rtctools/data/util.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
+from typing import List
+
+import numpy as np
+
+
+def check_times_are_increasing(times: List[datetime]):
+    """Check that time stamps are increasing."""
```
jackvreeken, line 8: Feels very LLM'y. No need for these docstrings imo (if they match the function name almost exactly)

src/rtctools/data/util.py
```diff
@@ -0,0 +1,41 @@
...
+    for i in range(len(times) - 1):
+        if times[i + 1] - times[i] != dt:
+            raise ValueError(
+                "Expecting equidistant timeseries, the time step towards "
+                "{} is not the same as the time step(s) before. ".format(times[i + 1])
+            )
+
+
+def fill_nan_in_timeseries(times: List[datetime], values: np.ndarray, interp_args: dict = None):
+    """Fill in missing values in a timeseries using lienar interpolation.
```
jackvreeken, line 26: Typo "linear"

src/rtctools/data/util.py
```diff
@@ -0,0 +1,41 @@
...
+    dt = times[1] - times[0]
+    for i in range(len(times) - 1):
+        if times[i + 1] - times[i] != dt:
+            raise ValueError(
+                "Expecting equidistant timeseries, the time step towards "
+                "{} is not the same as the time step(s) before. ".format(times[i + 1])
+            )
+
+
+def fill_nan_in_timeseries(times: List[datetime], values: np.ndarray, interp_args: dict = None):
```
jackvreeken, line 25: This new function is not used anywhere? So it does not feel like it belongs in a "refactor" commit. Also we wary of duplicating the `def __interpolate` code in `optimization_problem.py` .

tests/optimization/test_csv_mixin.py
```diff
@@ -112,3 +116,15 @@ def setUp(self):
     def test_objective_value(self):
         objective_value_tol = 1e-6
         self.assertTrue(abs(self.problem.objective_value) < objective_value_tol)
+
+
+class TestReadTimeseries(TestCase):
+    def test_get_timeseries_from_csv(self):
+        """Test getting a timeseries from a cvs file."""
```
jackvreeken, line 123: typo, csv

tests/data/test_util.py
```diff
@@ -0,0 +1,47 @@
...
+    fill_nan_in_timeseries,
+)
+from test_case import TestCase
+
+DATA_DIR = Path(__file__).parent / "data" / "timeseries"
+
+
+class TestUtil(TestCase):
+    def test_check_times_are_equidistant(self):
+        """Test checking if a timeseries is equidistant."""
```
jackvreeken, line 18: Same comment for all those docstrings. I personally don't feel they add much, but if you're really invested in them being there it's OK to keep them.

tests/data/test_util.py
```diff
@@ -0,0 +1,47 @@
...
+        """Test checking times are strictly increasing."""
+        times = [datetime(2020, 1, 1, 0, 0, sec) for sec in [1, 2, 3]]
+        check_times_are_increasing(times)
+        times = [datetime(2020, 1, 1, 0, 0, sec) for sec in [1, 2, 2]]
+        with self.assertRaises(ValueError):
+            check_times_are_increasing(times)
+
+    def test_fill_nan_in_timeseries(self):
+        """Test filling nan values of a timeseries."""
+        del self
```
jackvreeken, line 35: Ehhh, what is this?

jackvreeken (changes requested):
- Where is NetCDFMixin's use of "check_times_are_increasing" ?
- Wouldn't these changes also apply to simulation's csv_mixin / pi_mixin? And because you're making "data/util.py", I feel like it's optimization/simulation agnostic, so it _should_ work there as well.
- Wouldn't it be easier/cleaner to move the checks on monotonicity / equidistant to the IOMixin / DataStorage? Did you try to see what that solution would look like?

tests/data/test_util.py
SGeeversAtVortech: I'll remove this. I did this to indicate that parameter `self` is not used in this function, but we don't do this anywhere else in rtc-tools, so I'll just remove it.

tests/data/test_util.py
SGeeversAtVortech: It's a bit of a habit to add a docstring to each function/class. I know it does not always add much, but I think it doesn't hurt either.

src/rtctools/data/util.py
SGeeversAtVortech: Good point! It should definitely be in a separate commit. I'll see if I can reuse the code in `__interpolate`.

SGeeversAtVortech:
> * Where is NetCDFMixin's use of "check_times_are_increasing" ?
> * Wouldn't these changes also apply to simulation's csv_mixin / pi_mixin? And because you're making "data/util.py", I feel like it's optimization/simulation agnostic, so it _should_ work there as well.
> * Wouldn't it be easier/cleaner to move the checks on monotonicity / equidistant to the IOMixin / DataStorage? Did you try to see what that solution would look like?

Good points! I'll check if the functions introduced here can also be applied in simulation and NetCDFMixin. I would rather keep them as separate functions instead of moving them to an `IOMixin` so users can also use these functions in their own post/pre methods.

SGeeversAtVortech:
This was required for https://github.com/Deltares/rtc-tools/pull/1645, but we now use https://github.com/Deltares/rtc-tools/pull/1680 instead.

PR #1650, opened 2025-01-29
Add deploy to PyPI job

jackvreeken:
And some related fixes/improvements. Feel free to pick-and-choose commits if you don't like some of them


.github/workflows/rtc-tools.yml
```diff
@@ -88,22 +89,35 @@ jobs:
...
-  #   - uses: actions/checkout@v4.1.0
-  #   - uses: actions/download-artifact@v4.1.0
-  #     with:
-  #       name: "${{ github.job }}"
-  #   - run: pip install twine
-  #   - run: twine upload -u $PYPI_USER -p $PYPI_PASSWORD dist/*
+  deploy:
+    needs: [test-linux, coverage, examples-linux]
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags')  }}
```
SGeeversAtVortech, line 101: I would use here `startsWith(github.ref, 'refs/tags/2.')` instead of just `startsWith(github.ref, 'refs/tags)` to avoid every single tag will be pushed to pypi.

.github/workflows/rtc-tools.yml
```diff
@@ -88,22 +89,35 @@ jobs:
...
+    permissions:
+      id-token: write  # Required for OIDC PyPI publishing
+      contents: read
```
SGeeversAtVortech, line 104: I'm still not sure why my deploy job in the add_cd branch was not working, but perhaps it was because I was missing this line. Versioneer requires some information about the commits and tags to determine the version.

.github/workflows/rtc-tools.yml
```diff
@@ -88,22 +89,35 @@ jobs:
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
```
SGeeversAtVortech, line 112: Shouldn't we use a specific version here, like 3.9?

.github/workflows/rtc-tools.yml
```diff
@@ -88,22 +89,35 @@ jobs:
...
+
+      - name: Install build dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install build
+
+      - name: Build package
+        run: python -m build
```
SGeeversAtVortech, line 120: This command looks a lot nicer than the one in the build job. Are they doing the same thing? If so, I would update the build job.

SGeeversAtVortech (commented):
Thanks a lot for fixing the deploy job!

I left a few questions/remarks, but overall it looks good to me.

.github/workflows/rtc-tools.yml
jackvreeken: I think so yes, I'll add a commit to update the build job command.

.github/workflows/rtc-tools.yml
jackvreeken: Could be. Also, I could only really see lingering "v2.7.0a4" tags for example; those never triggered correctly because tags shouldn't start with a "v".

.github/workflows/rtc-tools.yml
SGeeversAtVortech: Ok, I think this is fine then.

PR #1651, opened 2025-02-04
Remove dependency on pkg resources

jackvreeken:
The pkg_resources package has been removed from the Python standard
library in version 3.12, while still available when (manually)
installing setuptools. A more modern approach is the usage of
"importlib", available since Python 3.9. Note that we do require a
change to "entry_point" accepting a group parameter that was introduced
with Python 3.10, so we fall back to a version of "importlib_metadata"
that support this. Also note that we also only require this external
(backport) dependency for Python 3.9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,6 @@ RTC-Tools uses [CasADi](https://web.casadi.org/) as a symbolic framework for alg
 pip install rtc-tools
 ```
 
-When using Python 3.12 or higher, you might first need to install setuptools
-
-```bash
-pip install setuptools
-```
-
 ## Documentation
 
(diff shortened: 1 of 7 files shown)


jackvreeken:
That was already in the commit message, it's just that the Github PR didn't pick up on it because I was building on top of the other PR.

jarsarasty:
> That was already in the commit message, it's just that the Github PR didn't pick up on it because I was building on top of the other PR. Copy pasted it in now.

Thanks, @jackvreeken

jackvreeken:
@SGeeversAtVortech any comments on this one?

PR #1652, opened 2025-02-04
Style switch to ruff format

jackvreeken:
- [x] Builds on #1650 and #1651 , so merge those first
- [ ] Add commit to .git-blame-ignore-revs (after merge)

### DIFF
--- a/tests/optimization/test_control_tree_mixin.py
+++ b/tests/optimization/test_control_tree_mixin.py
@@ -1,13 +1,14 @@
 import logging
 
 import numpy as np
+from test_case import TestCase
+
 from rtctools.optimization.collocated_integrated_optimization_problem import (
     CollocatedIntegratedOptimizationProblem,
 )
 from rtctools.optimization.control_tree_mixin import ControlTreeMixin
 from rtctools.optimization.modelica_mixin import ModelicaMixin
 from rtctools.optimization.timeseries import Timeseries
-from test_case import TestCase
 
 from .data_path import data_path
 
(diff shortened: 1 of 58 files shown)


tests/optimization/test_control_tree_mixin.py
```diff
@@ -1,13 +1,14 @@
 import logging
 
 import numpy as np
+from test_case import TestCase
```
SGeeversAtVortech, line 4: Ruff considers these relative imports like test_case as an external package?

SGeeversAtVortech (commented):
Just to be sure, most changes were generated by ruff except for the .pre-commit-config.yaml and pyproject.toml?

jackvreeken:
Yes, by far most changes are automatic. There were a few `==` checks that needed to be `is`, and a line or two that were too long and Ruff doesn't know how to wrap it.

tests/optimization/test_control_tree_mixin.py
jackvreeken: It's not a relative import like this, it would need to be `from .test_case` then. We can fix that, but as your other question rightfully suggests, mixing different types of fixes into this one commit is not a great idea.

SGeeversAtVortech (approved):
Seems like a nice improvement!

jackvreeken:
Who has to press the button? ;)

SGeeversAtVortech:
I'll do it :).

SGeeversAtVortech:
@jackvreeken: could you update the .git-blame-ignore-revs?

PR #1653, opened 2025-02-06
CSVMixin: Enable support for ensembles with one member

jbaayen:
Ensembles with one member failed to load previously, because np.genfromtxt() would return a 0-dimensional array when reading 'ensemble.csv'.  This change lifts the array if 0-dimensional, and adds a test case.


SGeeversAtVortech (approved):
Looks good to me.

PR #1656, opened 2025-02-13
PiMixin: Enable support for ignoring provided forecast dates

Ailbhemit:
An exception is raised if forecast dates were provided and they are not all equal. This option allows users to ignore provided forecast dates.

### DIFF
--- a/src/rtctools/data/pi.py
+++ b/src/rtctools/data/pi.py
@@ -350,6 +350,7 @@ class Timeseries:
         basename,
         binary=True,
         pi_validate_times=False,
+        pi_read_forecast_date=True,
         make_new_file=False,
     ):
         """
@@ -454,7 +455,7 @@ class Timeseries:
                     )
 
                 el = header.find("pi:forecastDate", ns)
-                if el is not None:
+                if el is not None and pi_read_forecast_date:
                     forecast_datetime = self.__parse_date_time(el)
                 else:
                     # the timeseries has no forecastDate, so the forecastDate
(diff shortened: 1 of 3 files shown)


src/rtctools/simulation/pi_mixin.py
```diff
@@ -41,6 +41,9 @@ class PIMixin(IOMixin):
     #: Check consistency of timeseries
     pi_validate_timeseries = True
 
+    #: Read forecast date from timeseries_import
+    pi_read_forecast_date = True
```
SGeeversAtVortech, line 45: Please describe this var in the class docstring.

SGeeversAtVortech (commented):
This option is basically equivalent to preprocessing xml files by removing all forecast_date elements. I would suggest to rename the option as ignore_forecast_date.

My main concern is if this is the correct place to apply preprocessing on xml files. Unfortunately, I don't know if there are alternative ways to achieve this through fews, fews-io, or some other library.

src/rtctools/data/pi.py
```diff
@@ -350,6 +350,7 @@ def __init__(
         basename,
         binary=True,
         pi_validate_times=False,
+        pi_read_forecast_date=True,
         make_new_file=False,
     ):
         """
```
jackvreeken, line 356: Also update the docstring

src/rtctools/data/pi.py
```diff
@@ -454,7 +455,7 @@ def __init__(
                     )
 
                 el = header.find("pi:forecastDate", ns)
-                if el is not None:
+                if el is not None and pi_read_forecast_date:
                     forecast_datetime = self.__parse_date_time(el)
                 else:
                     # the timeseries has no forecastDate, so the forecastDate
```
jackvreeken, line 461: Doesn't this mean that you will now just get an error if all start datetimes don't match? That doesn't seem like much of an improvement

jackvreeken (changes requested):

PR #1659, opened 2025-02-26
Updating exit flags of Knitro

FaridAlavi:


jackvreeken:
The bigger question is; do we want to support these codes for all solvers out of the box? If so, what codes do we implement? I vaguely remember having copy/paste code for e.g. CPLEX/Gurobi/XPRESS to detect "time limit exceeded" as succesful for MILPs, is that something we want to bring into the RTC-Tools core, or explicitly do _not_ want that in there?

src/rtctools/optimization/optimization_problem.py
```diff
@@ -313,6 +313,10 @@ def solver_success(
 
         if log_level == logging.ERROR and not log_solver_failure_as_error:
             log_level = logging.INFO
+        
+        if self.solver_options()["solver"].lower() in ["knitro"]:
+            if solver_stats["return_status"] in ["KN_RC_OPTIMAL_OR_SATISFACTORY", "KN_RC_ITER_LIMIT_FEAS"]:
+                success = True
```
jarsarasty, line 322: Should we also include KN_RC_NEAR_OPT as a success return status?

jarsarasty: If we consider KN_RC_ITER_LIMIT_FEAS to be a success, why not consider these other flags to be successes, too, if they all indicate feasible solutions?
- KN_RC_FEAS_XTOL
- KN_RC_FEAS_NO_IMPROVE
- KN_RC_FEAS_FTOL
- KN_RC_TIME_LIMIT_FEAS
- KN_RC_FEVAL_LIMIT_FEAS
- KN_RC_MIP_EXH_FEAS
- KN_RC_MIP_TERM_FEAS
- KN_RC_MIP_SOLVE_LIMIT_FEAS
- KN_RC_MIP_NODE_LIMIT_FEAS

jarsarasty (changes requested):
Hi @FaridAlavi 

Thanks for this pull request. It seems that only KN_RC_OPTIMAL_OR_SATISFACTORY and KN_RC_NEAR_OPT are interpreted as success in Casadi's Knitro interface. My impression is that other success states corresponding to feasible solutions also need to be overwritten.

jarsarasty:
> The bigger question is; do we want to support these codes for all solvers out of the box? If so, what codes do we implement? I vaguely remember having copy/paste code for e.g. CPLEX/Gurobi/XPRESS to detect "time limit exceeded" as succesful for MILPs, is that something we want to bring into the RTC-Tools core, or explicitly do _not_ want that in there?

Hi @jackvreeken 

Good point! Indeed, we need to double check that we are covering all the relevant cases for this solver. 

Regarding the bigger question: we would like to support only a few solvers "out of the box". We could argue for Kintro as a leading mixed-integer nonlinear optimization solver supported by Casadi.

src/rtctools/optimization/optimization_problem.py
FaridAlavi: Considering that in the applications of RTC-Tools, the best feasible solution is still useful, KN_RC_NEAR_OPT should also be considered a success flag, along with the list of other mentioned flags

jarsarasty:
Hi @FaridAlavi,

After updating the pip package in the workflow, the pipeline now works

FaridAlavi:
> Hi @FaridAlavi,
> 
> After updating the pip package in the workflow, the pipeline now works

Excellent! The branch is now ready for merge.

jarsarasty:
@FaridAlavi,

I cannot complete the merge.
I asked @SGeeversAtVortech to review again the changes and merge

src/rtctools/optimization/optimization_problem.py
```diff
@@ -314,6 +314,24 @@ def solver_success(
         if log_level == logging.ERROR and not log_solver_failure_as_error:
             log_level = logging.INFO
 
+        if self.solver_options()["solver"].lower() in ["knitro"]:
```
jackvreeken, line 317: Why not `== "knitro"` ?

src/rtctools/optimization/optimization_problem.py
```diff
@@ -314,6 +314,24 @@ def solver_success(
...
+            list_feas_flags = [
```
jackvreeken, line 318: Checking if an element exists is faster ( O(1) instead of O(N) ) if you make this a set `{...}`

jarsarasty:
Hi @FaridAlavi,

Github says that _this branch cannot be rebased due to conflicts_. Could you please resolve the conflicts before we can proceed with the merge?

jackvreeken:
Still could do with some more rebasing on master?

jackvreeken:
There's 11 commits, some of which are already on master. Also, the convention in RTC-Tools is to not have "fixup commits" in a PR. Those should be squashed into nicely formatted and commented atomic commits.

jarsarasty:
Thanks, @FaridAlavi!

The tests are passing now

PR #1661, opened 2025-03-04
Linear history

jackvreeken:
This needs to be force pushed to master to get rid of the merge commits.

Note that there'll still be merge commits from all the way back in 2017, but I don't want to rebase those away as we have made releases with those hashes.


SGeeversAtVortech:
The commit history looks correct. I'll do a force push this one time.

PR #1669, opened 2025-03-26
sim: warning should not be logged if default values were used

Ailbhemit:
A debug message is added instead of a warning if default values are used in initialization of simulation

### DIFF
--- a/src/rtctools/simulation/simulation_problem.py
+++ b/src/rtctools/simulation/simulation_problem.py
@@ -526,10 +526,14 @@ class SimulationProblem(DataStoreAccessor):
             # Check that the start_value is in between the variable bounds.
             if start_is_numeric and not min_is_symbolic and not max_is_symbolic:
                 if not (var.min <= start_val and start_val <= var.max):
-                    logger.warning(
-                        "Initialize: start value {} = {}".format(var_name, start_val)
-                        + " is not in between bounds {} and {}".format(var.min, var.max)
-                        + " and will be adjusted."
+                    logger.log(
+                        (
+                            logging.WARNING
+                            if source_description != "modelica file or default value"
+                            else logging.DEBUG
+                        ),
+                        f"Initialize: start value {var_name} = {start_val} "
+                        f"is not in between bounds {var.min} and {var.max} and will be adjusted.",
                     )
 
         # Default start var for ders is zero


src/rtctools/simulation/simulation_problem.py
```diff
@@ -526,10 +526,14 @@ def initialize(self, config_file=None):
...
                 if not (var.min <= start_val and start_val <= var.max):
-                    logger.warning(
-                        "Initialize: start value {} = {}".format(var_name, start_val)
-                        + " is not in between bounds {} and {}".format(var.min, var.max)
-                        + " and will be adjusted."
+                    logger.log(
+                        (
+                            logging.WARNING
+                            if source_description != "modelica file or default value"
+                            else logging.DEBUG
```
SGeeversAtVortech, line 533: Why not give a warning? The default value is outside the given bounds, so it seems a bit of a strange default value. It makes sense to me to give a warning in this case...

Ailbhemit: The user doesn't provide a default value, RTC-Tools does, so a warning message does not make sense if an internally defined value is overwritten to be within bounds. Of course a warning message can still be given in the other cases where the user has provided the value.

SGeeversAtVortech: Ah yes, if the user does not provide any initial value, it seems pymoca sets a the start value to 0. This might be outside the bounds of the variable, so you do not want an error messsage in this case.

SGeeversAtVortech (approved):

SGeeversAtVortech:
@jackvreeken If you're ok with this PR I'd like to merge it. It only changes a warning messages to a debug message.

jackvreeken:
I got very confused about the `fixed=True` case, as the "Check" is outside the if-else that's right above it. The check only makes sense for the "else" case however. 

So the diff is OK, but I would change the commit message a little because it's somewhat vague right now. And also because every time it takes me quite a while to understand all the stuff that is going on in SimulationProblem 😅

```
sim: Do not warn out-of-bounds start value if not explicitly set

We would warn users about initial/start values they never explicitly
set. The log message is specifically about states that are minimized
(not constrained/bounded), so 'out-of-bounds' start values would not
lead to infeasible problems. To not needlessly confuse the user with
warnings, we change the message to "DEBUG" level, unless they explicitly
set a start value themselves in which case we keep the "WARNING" level.

Note that there is no good way (yet) to separate a default value of 0
coming from Pymoca from an explicitly specified value of 0. The
assumption (unchanged in this commit) is that 0 means 'default'.
```


I did look into the "start value default" thing in Pymoca, and it seems that internally at least Pymoca keeps track of this by having a `_DefaultValue` class (inheriting from int). So that might be a way to separate "explicitly specified 0" from "default value 0". I only checked the Pymoca side of things, but I think that that `_DefaultValue`-type persists even for cached models. Definitely not something for _this_ PR though, maybe some future work (on both the Pymoca side and RTC-Tools side).

jackvreeken (changes requested):
(just the commit message, to clarify/elaborate)

jackvreeken (approved):

PR #1672, opened 2025-04-02
CollInt: Fix __debug_check_state_output_scaling for ensemble members

jackvreeken:
The ensemble indices are lists. We already converted the the first ensemble member's to a set, but not the remaining ones. Trying to do `|=` with a list on the right-hand side is not allowed, so we have to explicitly converted to a set first.


jackvreeken:
Good catch! Fixed :)

SGeeversAtVortech (approved):

PR #1678, opened 2025-04-10
setup: Update to Casadi 3.7.0

jarsarasty:
- Updated setup.py to use Casadi 3.7.0
- Updated requirements.txt to reflect the new Casadi version


setup.py
```diff
@@ -45,7 +45,7 @@
     packages=find_packages("src"),
     package_dir={"": "src"},
     install_requires=[
-        "casadi >= 3.6.3, == 3.6.*, !=3.6.6",
+        "casadi >= 3.6.3, <= 3.7.0, !=3.6.6",
```
SGeeversAtVortech, line 48: Why not just `<=3.7`?

jarsarasty: Hi @SGeeversAtVortech 

The idea was to be explicit about the maximum Casadi version, since Casadi does not strictly follow semantic versioning. We could relax it to 3.7 if there is a good reason for that.

SGeeversAtVortech: Bumps in casadi major versions sometimes introduce breaking changes, but I don't expect breaking changes in patches.

jarsarasty: Changed

doc/requirements.txt
```diff
@@ -1,4 +1,4 @@
-casadi >= 3.6.3, == 3.6.*
+casadi >= 3.6.3, <= 3.7.0, !=3.6.6
```
SGeeversAtVortech, line 1: Match the requirement with setup.py?

jarsarasty: It's updated now

SGeeversAtVortech (approved):

PR #1682, opened 2025-05-06
Optimization: update suggested inheritance order

SGeeversAtVortech:
Update the suggested inheritance order for an optimization problem. In particular, add IOMixin to the list.
If GoalProgrammingMixin and IOMixin are interchanged, for example, this can lead to unexpected results/errors.

### DIFF
--- a/src/rtctools/util.py
+++ b/src/rtctools/util.py
@@ -110,6 +110,7 @@ def run_optimization_problem(
             "GoalProgrammingMixin",
             "PIMixin",
             "CSVMixin",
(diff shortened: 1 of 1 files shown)


jackvreeken:
IOMixin is not really a class that people import themselves, but I assume you ran into this issue with a custom IO class?

SGeeversAtVortech:
Indeed, I used a sort of customized IO class that wrote/read timeseries, but never really saved them to files. The timeseries were passed to/from a SimulationProblem class.

PR #1687, opened 2025-05-20
ControlTreeMixin: Assign one ensemble member per branch when possible

jbaayen:
When constructing the control tree and the number of branches k is greater than or equal to the number of available ensemble members, assign one member to each branch.

This change ensures a more intuitive distribution: when the tree is large enough, each branch receives at most one ensemble member. It avoids scenarios where multiple members are assigned to the same branch while others remain empty.

The updated behavior is verified with two new tests.


jarsarasty:
Hi @jbaayen ,

As discussed yesterday, this PR will be put on hold for now.

PR #1689, opened 2025-05-20
Fix rtc-tools-download-examples script

jackvreeken:
On GitLab, the zipfile of the tree that was downloaded contained a folder with the tag name in it. For GitHub however, the folder is names '<group>-<repo>-<sha>', e.g. "Deltares-rtc-tools-1c7b20a". We therefore just find a folder named like that instead.


jarsarasty:
Thanks, @jackvreeken

PR #1693, opened 2025-05-31
Bounds per ensemble member (feature flag, backwards compatible)

jackvreeken:


src/rtctools/optimization/goal_programming_mixin_base.py
```diff
@@ -394,7 +394,17 @@ def __init__(self, optimization_problem):
         # Extract state range from model
         if self.has_target_bounds:
             try:
-                self.function_range = optimization_problem.bounds()[self.state]
+                # Handle ensemble-specific bounds by using ensemble_member=0 if
+                # feature flag is enabled
+                # TODO: What do we do with this? Max? Min? User explicit input?
```
jbaayen, line 399: Why don't we check here if the bounds for this variable are the same for all members. If they are, we proceed as usual. If they are not, we raise an exception and instruct the caller to set `function_range` manually.

src/rtctools/optimization/optimization_problem.py
```diff
@@ -221,7 +225,14 @@ def optimize(
     def __check_bounds_control_input(self) -> None:
         # Checks if at the control inputs have bounds, log warning when a control input is not
         # bounded.
-        bounds = self.bounds()
+        if self.ensemble_specific_bounds:
+            # We can just naively merge dictionaries, as we're just checking for
+            # presence of keys.
```
jbaayen, line 230: What if bounds are missing for a control variable in one of the members? Wouldn't we want to raise the warning in that case?

src/rtctools/optimization/collocated_integrated_optimization_problem.py
```diff
@@ -2043,7 +2047,67 @@ def integrator_options(self):
...
+            # take the minimum.
+            ensemble_lbx_ubx = np.array(
+                [
+                    self._collint_get_lbx_ubx_member(bounds[i], count, indices)
+                    for i in range(self.ensemble_size)
+                ]
+            )
+
+            lbx = np.max(ensemble_lbx_ubx[:, 0, :], axis=0)
+            ubx = np.min(ensemble_lbx_ubx[:, 1, :], axis=0)
```
jbaayen, line 2071: Should this logic only be applied to control variables?

jbaayen (changes requested):

src/rtctools/optimization/goal_programming_mixin_base.py
jackvreeken: Will still be a somewhat ugly check (support np.ndarray, Timeseries, and floats), but could work yes.

src/rtctools/optimization/collocated_integrated_optimization_problem.py
jbaayen: But isn't this very same function also being called from `discretize_states()`?

src/rtctools/optimization/goal_programming_mixin_base.py
jackvreeken: Made something anyway; the user will just have to explicitly set something and then override it later I guess (if they're feeling fancy)

jarsarasty (approved):
Thanks, @jackvreeken !

The implementation of this feature look good to me, and Jorn's comments have been addressed.

jarsarasty:
Thanks, @jackvreeken!

Please let me know if you need a dev release

PR #1694, opened 2025-06-01
CollInt: Support returning times from 'states_in()'

jackvreeken:
Internally the `times` were already returned with the `__states_time_in` method. Exposing this functinality to the user will allow them to e.g. compute integrals/sums of algebraic states/control inputs.

Our implementation of `integral` takes averages over a time step, which makes sense for state variables like reservoir volumes. But if one is interested in a sum/average of algebraic states, they would want to exclude the value at t0 (assuming backwards euler). This is non-trivial, and would mean the user has to reimplent the logic in `__states_time_in` for history etc.

Note that we may want to support different ways of calculating the integral directly in `integral()` in the future, but at least optionally returning the times from `states_in()` in this commit already allows this and more in a generic way.

### DIFF
--- a/src/rtctools/optimization/collocated_integrated_optimization_problem.py
+++ b/src/rtctools/optimization/collocated_integrated_optimization_problem.py
@@ -2639,10 +2639,21 @@ class CollocatedIntegratedOptimizationProblem(OptimizationProblem, metaclass=ABC
 
         return x, t
 
-    def states_in(self, variable, t0=None, tf=None, ensemble_member=0):
-        x, _ = self.__states_times_in(variable, t0, tf, ensemble_member)
+    def states_in(
+        self,
+        variable: str,
+        t0: float = None,
+        tf: float = None,
+        ensemble_member: int = 0,
+        *,
+        return_times: bool = False,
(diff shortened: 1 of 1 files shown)


src/rtctools/optimization/collocated_integrated_optimization_problem.py
```diff
@@ -2612,10 +2612,13 @@ def __states_times_in(self, variable, t0=None, tf=None, ensemble_member=0):
 
         return x, t
 
-    def states_in(self, variable, t0=None, tf=None, ensemble_member=0):
-        x, _ = self.__states_times_in(variable, t0, tf, ensemble_member)
+    def states_in(self, variable, t0=None, tf=None, ensemble_member=0, *, return_times=False):
```
jarsarasty, line 2615: I would recommend including type hints whenever we introduce changes to a function.

jarsarasty (commented):
Indeed, this is a useful feature. Thanks, @jackvreeken!

I only made one suggestion regarding the use of type hints.

jackvreeken:
Yep, definitely good to be adding those when we can! Too bad the `|` operator is needs Python 3.10 and I still have to use `Union`, but we can "fix" that later this year when 3.9 gets EOL.

PR #1696, opened 2025-06-03
Test with oldest and newest versions of deps

jackvreeken:
Needs #1695


jackvreeken:
For sure! It builds on the #1695 one, so when that PR is merged, this one can move to review.

jarsarasty:
@jackvreeken,
Now that #1695 has been merged, could you please update this PR?

.github/workflows/rtc-tools.yml
```diff
@@ -64,6 +64,40 @@ jobs:
         dependency-groups: "tests"
     - run: uv run tox -vv
 
+  test-linux-oldest-deps:
```
jarsarasty, line 67: To avoid code duplication, would you consider a workflow matrix for this part?

pyproject.toml
```diff
@@ -32,8 +32,8 @@ classifiers = [
 ]
 dependencies = [
     "casadi >= 3.6.3, <= 3.7, !=3.6.6",
-    "numpy >= 1.16.0",
-    "scipy >= 1.0.0",
+    "numpy >= 1.20.0",
```
jarsarasty, line 35: The new testing setup with the oldest and newest versions of the dependencies will catch breakages during testing in the CI pipeline. However, if a breaking dependency update is released after the last CI run, users installing the package may get an incompatible version.
Should we pin the upper bound of these dependencies?

jarsarasty (changes requested):
Hi @jackvreeken

Thanks for these improvements for the tests.

It looks good! I just made two small comments. Please let me know what you think.

jackvreeken:
- Add running every week.
- Also run _examples_  (every week) with new/old version

.github/workflows/rtc-tools.yml
jackvreeken: Yep, will merge newest + oldest.

jackvreeken:
@jarsarasty works now!

jarsarasty:
@jackvreeken 
Please update the commit message to clearly indicate the required versions: Bump minimum required versions to numpy ≥ 1.19.0 and scipy ≥ 1.5.0 for Python 3.9 wheel support

PR #1699, opened 2025-06-08
CollInt/_debug_*: Fix tol_down warnings in coefficient check

jackvreeken:
- [x] Needs test; e.g. a coefficient of 1e-6 and 0.0, setting tol_down to 1e-12. It would trigger on the 0.0, which this commit fixes.

### DIFF
--- a/src/rtctools/optimization/collocated_integrated_optimization_problem.py
+++ b/src/rtctools/optimization/collocated_integrated_optimization_problem.py
@@ -2947,8 +2947,19 @@ class CollocatedIntegratedOptimizationProblem(OptimizationProblem, metaclass=ABC
             "{} & {}, {} & {}".format(max_constr_A, min_constr_A, max_constr_b, min_constr_b)
         )
 
-        maxs = [x for x in [max_constr_A, max_constr_b, max_obj_A, obj_b] if x is not None]
-        mins = [x for x in [min_constr_A, min_constr_b, min_obj_A, obj_b] if x is not None]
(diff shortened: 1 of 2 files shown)


jackvreeken:
SonarCloud complaining that I shouldn't compare floats to 0.0. In general an OK suggestion, but not for this commit. For solvers, _exactly_ zero is fine, we care about small **non-zero** values.

src/rtctools/optimization/collocated_integrated_optimization_problem.py
```diff
@@ -2920,8 +2920,12 @@ def constr_to_str(i):
...
         )
 
-        maxs = [x for x in [max_constr_A, max_constr_b, max_obj_A, obj_b] if x is not None]
-        mins = [x for x in [min_constr_A, min_constr_b, min_obj_A, obj_b] if x is not None]
+        maxs = [
+            x for x in [max_constr_A, max_constr_b, max_obj_A, obj_b] if x is not None and x != 0.0
+        ]
+        mins = [
+            x for x in [min_constr_A, min_constr_b, min_obj_A, obj_b] if x is not None and x != 0.0
+        ]
```
jarsarasty, line 2928: We could check if inline supression (# NOSONAR) works

A comment could be added to indicate that the direct comparison to 0.0 is intentional.

jarsarasty: Thanks!

jarsarasty (approved):

PR #1703, opened 2025-06-26
Correct previously introduced wrong solver flag: when the run failes …

k-horvath-deltares:
…the message was stating success.


jarsarasty (changes requested):

jarsarasty (approved):
Hi @k-horvath-deltares ,

Thanks for addressing the comments! It looks good to me!

PR #1704, opened 2025-07-04
CollocatedIntegratedOptimizationProblem:  Add option to disable initial constraints

jbaayen:
Initial constraints are only required when an incomplete set of initial states is provided. If the set of initial states is either complete or "sufficiently" complete (e.g., values of all differentiated states are provided), then initial constraints may or may not be required depending on the use case.

This change introduces an option that controls explicitly whether or not the initial constraints are created.

### DIFF
--- a/src/rtctools/optimization/collocated_integrated_optimization_problem.py
+++ b/src/rtctools/optimization/collocated_integrated_optimization_problem.py
@@ -44,10 +44,16 @@ class CollocatedIntegratedOptimizationProblem(OptimizationProblem, metaclass=ABC
 
     :cvar check_collocation_linearity:
         If ``True``, check whether collocation constraints are linear. Default is ``True``.
+    :cvar add_initial_constraints:
+        If ``True``, add constraints for initial conditions. Default is ``True``.
+        Set to ``False`` when all initial states are provided to avoid an overdetermined system.
     """
 
     #: Check whether the collocation constraints are linear
     check_collocation_linearity = True
+
+    #: Add constraints for initial conditions
+    add_initial_constraints = True
 
     #: Whether or not the collocation constraints are linear (affine)
     linear_collocation = None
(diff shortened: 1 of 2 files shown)


jarsarasty:
@jbaayen 
It appears that a few conflicts need to be resolved in this branch.

jarsarasty (changes requested):
Hi @jbaayen ,

Thank you for this pull request, which extends the flexibility of the model specification in RTC-Tools.

Please consider adding validation to ensure that when `add_initial_constraints=False`, all initial states must be fully specified. 
When add_initial_constraints=True and all initial states are specified, a warning could be issue to inform the user that the system may be overdetermined.

src/rtctools/optimization/collocated_integrated_optimization_problem.py
```diff
@@ -44,11 +44,17 @@ class CollocatedIntegratedOptimizationProblem(OptimizationProblem, metaclass=ABC
 
     :cvar check_collocation_linearity:
         If ``True``, check whether collocation constraints are linear. Default is ``True``.
+    :cvar add_initial_constraints:
+        If ``True``, add constraints for initial conditions. Default is ``True``.
```
Ailbhemit, line 48: Could it also be made explicit to which variables this refers to: both constant and controlled inputs?

FaridAlavi:
After resolving the conflict, this merge can be completed.

jbaayen:
I now no longer think this new functionality should be merged.

Here is my rationale:
1. The standard workflow is to provide initial values for (differentiated) states and, perhaps, for control. In this workflow, the initial equations are, in fact, required.
2. Only when specifying the algebraic states as well, are the initial equations not needed. But if the algebraic state values provided are rounded (e.g. due to having been serialized and de-serialized), then this may result in an infeasible initial state.
3. We may not want to offer this new numerically "unsafe" workflow given that the standard workflow works well.

What do others think?

PR #1707, opened 2025-07-18
ControlTreeMixin:  Use int64 rather than int16 for control indices

jbaayen:
Nowhere else in the code do we use int16, and its use here was causing overflows for larger problems.


jarsarasty (approved):
Hi @jbaayen 
Thanks for this fix! int32 might have been sufficient in most cases, but int64 is definitely safer.

PR #1709, opened 2025-09-03
Initial guess for simulation is non-zero.

k-horvath-deltares:
For the initializaiton of simulaiton, the guess of solving the optimization problem was zero. However, when some veriables are in the denominator this resulted in problems for the solver. Starting from a non-zero initial guess the solver can find easier the root.


k-horvath-deltares:
This is a case-specific problem that can be solved by giving non-zero start values in Modelica for the specific variables.

PR #1712, opened 2025-09-29
CollocatedIntegratedOptimizationProblem: Add option to inline delay expressions

jbaayen:
As previously, by default, delay expressions are handled by introducing placeholder variables and by adding equality constraints equating the placeholder variables to the delayed expressions.

The default approach, however, introduces additional optimization variables and additional constraints.  For situations where this is not desirable, this commit introduces a new option
('inline_delay_expressions') that controls whether or not the delayed expressions are substituted into the discretized problem instead.

### DIFF
--- a/tests/optimization/test_delay_expressions.py
+++ b/tests/optimization/test_delay_expressions.py
@@ -17,13 +17,14 @@ logger.setLevel(logging.DEBUG)
 
 
 class Model(ModelicaMixin, CollocatedIntegratedOptimizationProblem):
-    def __init__(self):
+    def __init__(self, inline_delay_expressions=False):
         super().__init__(
             input_folder=data_path(),
             output_folder=data_path(),
             model_name="ModelDelay",
             model_folder=data_path(),
         )
+        self.inline_delay_expressions = inline_delay_expressions
 
     def times(self, variable=None):
         # Collocation points
@@ -63,7 +64,8 @@ class ModelCompleteHistory(Model):
 
 class TestDelayHistoryWarnings(TestCase, unittest.TestCase):
     def test_no_history(self):
-        problem = ModelNoHistory()
+        # Test default mode
+        problem = ModelNoHistory(inline_delay_expressions=False)
         with self.assertLogs(logger, level="WARN") as cm:
             problem.optimize()
             self.assertEqual(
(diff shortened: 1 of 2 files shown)


tests/optimization/test_delay_expressions.py
```diff
@@ -75,6 +77,26 @@ def test_no_history(self):
                     "Extrapolating t0 value backwards in time.",
                 ],
             )
+        results = problem.extract_results()
+
+        # Test inline mode
+        problem_inline = ModelNoHistory(inline_delay_expressions=True)
```
jackvreeken, line 83: Test succeeds with `inline_delay_expressions=False` as well

jbaayen:
> Also might be worthwhile explaining why you keep the variables around. I assume most solvers will handle `lbx == ubx`, but is it a deliberate choice over dropping the variables entirely?

Good point. I've added an explanation to the comments in the code (where I set lbx=ubx for the eliminated variables).

tests/optimization/test_delay_expressions.py
jackvreeken: I'd say to check the _pymoca_delay value in the results, based on the current approach. It should be zero for `inline_delay_expressions=True` right? And non-zero for `inline_delay_expressions=False`

jackvreeken:
Extra comment is great. Can you move it to the commit message instead though? The comment covers not just that section of the code, but the entire approach (and trade-offs in which you chose), so I feel the commit message is a more appropriate place.

tests/optimization/test_delay_expressions.py
jbaayen: I feel that approach would be exposing internals (in the form of private variables) to the tests, which I'm not convinced is a very clean approach?

jbaayen:
> Extra comment is great. Can you move it to the commit message instead though? The comment covers not just that section of the code, but the entire approach (and trade-offs in which you chose), so I feel the commit message is a more appropriate place.

I reckon both places would be best :).  I've added it to the commit message now as well.

tests/optimization/test_delay_expressions.py
jackvreeken: I'm not sure I follow. The whole idea of this PR is to change the behavior of those variables no? Then we should check that we do it correctly. 

Their names being prefixed with an underscore is more to indicate "don't rely on them being named exactly like this", but their presence very much can be relied on. Also, note that it is _Pymoca_ making them and naming them, not RTC-Tools.

You can just get their names via `[x[1] for x in problem_inline.delayed_feedback()]`.

jbaayen: I've added a check that the inlined problem constructs fewer constraints than the non-inlined problem.

PR #1720, opened 2025-10-21
setup.py: Fix allowing CasADi 3.7.*

jarsarasty:
For the maintenance 2.7, the Casadi version change is replicated from #1708:
A `<= 3.7` is equivalent to `<= 3.7.0`. We want to allow the entire 3.7 series of CasADi, and with `<= 3.7.*` not being allowed, we use `< 3.8`.


jackvreeken:
Ah, it's the maintenance branch, missed that :)

PR #1721, opened 2025-10-24
pi: Check assumption of zero-based ensembleMemberIndex.

Ailbhemit:
Explicitly check the existing assumption of zero-based ensembleMemberIndex. Raise a value error if this is not the case.

Relax this assumption if only a single ensembleMemberIndex is present. This allows more flexibility and the ability to execute model runs where a single non-zero ensembleMemberIndex is present.

### DIFF
--- a/src/rtctools/data/pi.py
+++ b/src/rtctools/data/pi.py
@@ -411,6 +411,7 @@ class Timeseries:
             self.__forecast_index = None
             self.__contains_ensemble = False
             self.__ensemble_size = 1
+            ensemble_indexes = set()
             for series in self.__xml_root.findall("pi:series", ns):
                 header = series.find("pi:header", ns)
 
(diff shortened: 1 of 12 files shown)


src/rtctools/data/pi.py
```diff
@@ -538,24 +548,27 @@ def __init__(
...
+                    el = header.find("pi:ensembleMemberIndex", ns)
+                    if el is not None:
+                        ensemble_member = int(el.text)
+                        while ensemble_member >= len(self.__values):
+                            self.__values.append({})
+                        while ensemble_member >= len(self.__units):
+                            self.__units.append({})
+                    else:
+                        ensemble_member = 0
+                    if el is None and self.contains_ensemble is True:
```
jarsarasty:
```suggestion
                    if el is None:
```

PR #1723, opened 2025-11-03
SinglePassGPMixin: Fix seeding of epsilons for the first priority

jackvreeken:
The seed() method for SinglePassGPMixin just returned the seed of GPMixinBase, but that one does not contain the seed of the epsilons that we add to the optimization problem.  Similar to "regular" GoalProgrammingMixin, we need to make sure to seed the epsilons to 1.0 for the first priority. We therefore copy-paste the relevant block over to SinglePassGPMixin.

Note that the `seed()` method for SinglePassGPMixin is only called for the first priority. Further priorities are handled by using the solver output of the previous priority as the seed.

### DIFF
--- a/tests/optimization/test_single_pass_goal_programming.py
+++ b/tests/optimization/test_single_pass_goal_programming.py
@@ -66,6 +66,7 @@ class ModelPathGoals:
         self._state_vectors = []
         self._solver_outputs = []
         self._constraints = []
+        self._initial_seed = None
 
     def times(self, variable=None):
         # Collocation points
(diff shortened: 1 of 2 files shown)


tests/optimization/test_single_pass_goal_programming.py
```diff
@@ -217,6 +223,17 @@ def test_seed(self):
...
 
+    def test_initial_seed(self):
+        """
+        Every goal's epsilon variable should be present in the initial seed, and
+        should be set to one.
+        """
+        n_goals = len(self.problem_append.path_goals()) + len(self.problem_append.goals())
+        self.assertEqual(len(self.problem_append._initial_seed.keys()), n_goals)
+
+        for ts in self.problem_append._initial_seed.values():
```
jbaayen: This is checking all seed values, not just for the goal programming epsilons

jarsarasty:
Hi @jackvreeken,

I had pushed a commit to the wrong branch. The branch has been reset to your latest commit.

PR #1735, opened 2025-11-27
solver_utils: add custom solver library preloading support

jarsarasty:
Implement mechanism to use custom versions of solver libraries instead of those bundled with CasADi. This enables users to leverage newer solver versions, apply patches, or use optimized builds.

Core implementation:
- Add solver_utils.py with thread-safe preloading via ctypes.CDLL
- Support explicit preloading and environment variable (RTCTOOLS_PRELOAD_SOLVER_LIBS)
- Automatic preloading on rtctools import when env var is set
- Import order validation and diagnostic utilities

Features:
- Platform-specific handling (RTLD_GLOBAL on Unix, winmode=0 on Windows)
- Thread-safe with optimized lock pattern (CDLL loading outside lock)
- Graceful degradation when preloading fails
- Comprehensive error messages with troubleshooting guidance

Testing:
- Currently supported and tested: HiGHS
- Comprehensive unit tests
- Separate integration test script for fresh-process testing
- Thread-safety tests covering concurrent preloading scenarios
- Configurable test library paths

Documentation:
- Complete user guide in doc/advanced/custom_solver_versions.rst
- Platform-specific instructions (Linux, macOS, Windows)
- Troubleshooting section with common issues
- Explicit IPOPT Windows limitation documented
- API reference with usage examples

Code quality improvements:
- Module-specific logger (__name__) for better filtering
- Simplified docstring examples
- Documented architecture and design decisions

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -1,0 +1,9 @@
+Advanced Topics
+===============
+
+This section covers advanced RTC-Tools features and configurations.
+
+.. toctree::
+   :maxdepth: 2
+
+   advanced/custom_solver_versions
--- a/doc/advanced/custom_solver_versions.rst
+++ b/doc/advanced/custom_solver_versions.rst
@@ -1,0 +1,219 @@
+Using Custom Solver Versions
+=============================
+
+By default, RTC-Tools uses the solver libraries bundled with CasADi. However, you may need to use different versions of solvers like HiGHS, IPOPT, or others for various reasons:
+
+* **Performance improvements** in newer solver releases
+* **Bug fixes** available in specific versions
+* **New features** not yet in CasADi's bundled version
+* **Custom builds** optimized for your hardware or compiled with specific options
+* **Version consistency** across development and production environments
+
+This guide explains how to use custom solver library versions with RTC-Tools.
+
+.. warning::
+   Custom solver libraries must be ABI-compatible with CasADi's expectations. Using incompatible builds may cause crashes or unexpected behavior.
+
+How It Works
+------------
+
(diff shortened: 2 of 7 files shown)


jackvreeken:
The standard way on Linux of overruling is something like 

```
LD_PRELOAD=/path/to/newer/libhighs.so.1 uv run pytest -s -k TestSolverHiGHS_MIP
```

That works equally well as `RTCTOOLS_PRELOAD_SOLVER_LIBS=highs:/home/pietje/Downloads/highs-v1.12.0-manylinux2014_x86_64/lib64/libhighs.so.1 uv run pytest -s -k TestSolverHiGHS_MIP`. On my system they _both_ load v1.12, but also both fail with a segfault. Might be because my local libhighs.so.1 has incompatibilities? What version of Highs are you testing with?

On MacOS I think `DYLD_INSERT_LIBRARIES= ...` should work equally well. So that leaves Windows, where the ctypes preloading functionality would be needed. Not sure it warrants more than a README/documentation entry though, as it essentially boils down to

```Python
ctypes.CDLL(r"C:\path\to\highs.dll")

import casadi
```

jackvreeken:
One question I have for @jgillis is, why is  `CASADIPATH` not first in the lookup order; why is it `casadi.GlobalOptions.getCasadiPath` ? The latter is set to the CasADi library folder by default, which makes it impossible to overrule bundled solvers like HiGHs. Is this by design?

Aside from that, and to also support newer version of libraries / solvers, I'd say we forego the whole OS-level interjecting. I think I prefer CASADIPATH overruling, but if that is not possible, something like:

```Python
import casadi
import os

casadi.GlobalOptions.setCasadiPath(f"/path/to/highs/and/its/casadi/plugin{os.pathsep}{casadi.GlobalOptions.getCasadiPath()}")
```

Might be convinced to support our own env var for this, but have to sleep on that a couple nights.

PR #1743, opened 2026-01-12
Fixes related to lowest/highest dependency resolution (NumPy 2.4 error, CasADi 3.6 substitution)

jackvreeken:


src/rtctools/_internal/casadi_helpers.py
```diff
@@ -53,9 +53,20 @@ def substitute_in_external(
     elif not expr:
         return []
     else:
+        # CasADi < 3.7: f.call() with MX values returns wrapped results like f(...){0}.
+        # Resolve symbolics with ca.substitute(), and convert resulting MX constants to
+        # floats first.
```
jarsarasty:
```suggestion
```

PR #1747, opened 2026-01-28
fix(CollInt): Correctly size input to delayed-feedback function in ca…

jbaayen:
…se of vector extra variables


jackvreeken:
See #1746

PR #1766, opened 2026-03-13
Examples: modified the cascading_channels Example.mo

BernhardBecker:
Modified the cascading_channels example, such that it fits to the new object in the channel flow library.

### DIFF
--- a/examples/cascading_channels/model/Example.mo
+++ b/examples/cascading_channels/model/Example.mo
@@ -55,9 +55,9 @@ model Example
   input Real DrinkingWaterExtractionPump_Q(fixed = false, min = 0., max = 2.) = DrinkingWaterExtractionPump.Q;
 initial equation
   // Steady state initialization
-  der(LowerChannel.Q) = 0.0;
-  der(MiddleChannel.Q) = 0.0;
-  der(UpperChannel.Q) = 0.0;
+  der(LowerChannel.Q) = 0;
+  der(MiddleChannel.Q) = 0;
+  der(UpperChannel.Q) = 0;
 equation
   connect(DrinkingWaterExtractionPump.HQDown, DrinkingWaterPlant.HQ) annotation(Line(points = {{20, 26}, {38, 26}, {38, 36}, {38, 36}}, color = {0, 0, 255}));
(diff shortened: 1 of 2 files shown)


BernhardBecker:
Question: is this correct? Why is this initialization three-dimensional? 
  // Steady state initialization
  der(LowerChannel.Q) = {0.0,0.0,0.0};
  der(MiddleChannel.Q) = {0.0,0.0,0.0};
  der(UpperChannel.Q) = {0.0,0.0,0.0};

examples/cascading_channels/model/Example.mo
```diff
@@ -1,70 +1,70 @@
...
+  parameter Real theta;
+  parameter Real step_size;
+  // Inputs
+  input Real Inflow_Q(fixed = true) = Inflow.Q;
+  input Real UpperControlStructure_Q(fixed = false, min = 0., max = 4.) = UpperControlStructure.Q;
+  input Real LowerControlStructure_Q(fixed = false, min = 0., max = 4.) = LowerControlStructure.Q;
+  input Real DrinkingWaterExtractionPump_Q(fixed = false, min = 0., max = 2.) = DrinkingWaterExtractionPump.Q;
+initial equation
+  // Steady state initialization
+  der(LowerChannel.Q) = {0.0,0.0,0.0};
```
k-horvath-deltares: Why do we need this form now? It is working with the old declaration. Where are the three numbers standing for and where are they used?

jarsarasty (commented):
Please see the updated comment below.

BernhardBecker:
The three-dimensional initialization is no longer needed. This addresses item 4 and item 1. Note that we can't validate the model in OMEdit, but this is due to a Pymoca issue. We should not wait until that one has been resolved.

jarsarasty:
When trying to run this example locally by checking out the modified `Example.mo`, it fails because `Deltares.ChannelFlow.Hydraulic.Branches.HomotopicRectangular` is not available in any published version of `rtc-tools-channel-flow` (latest is 1.3.1, and 1.4.0b2 also does not include it).

The class was added to the [Deltares/rtc-tools-channel-flow](https://github.com/Deltares/rtc-tools-channel-flow) repository on March 12, 2026 (commit by @k-horvath-deltares but has not been released to PyPI yet.

This PR should be kept open until a new release of `rtc-tools-channel-flow` that includes `HomotopicRectangular` is published.

PR #1767, opened 2026-03-16
feat(util): add RTCTOOLS_EXTRA_CASADIPATH for custom solver plugin path

jarsarasty:
CasADi searches for solver plugins in CASADIPATH but only after its own installation directory, making it impossible to override bundled solvers using CASADIPATH alone.

This PR adds `RTCTOOLS_EXTRA_CASADIPATH`, an environment variable that prepends a custom directory to CasADi's plugin search path via `GlobalOptions.setCasadiPath()`. This allows users to use custom-built CasADi plugin wrappers (e.g. libcasadi_conic_highs) compiled against a different solver version than the one bundled with CasADi.

## Implementation
- `_configure_extra_casadi_path(logger)` called after logging setup in both `run_optimization_problem()` and `run_simulation_problem()`
- Deduplication via `os.path.normcase(os.path.normpath(...))` — handles trailing-slash variants and case differences on Windows
- `logger.warning()` for each path entry that does not exist on disk
- `logger.info()` logs the active search path so users can confirm the override is in effect
- 6 unit tests
- Documentation added under Optimization > Solver Configuration

## Limitations
The directory must contain a CasADi plugin wrapper, not just the raw solver library. The wrapper must be compiled with the same toolchain as CasADi and linked against the desired solver version. See documentation for details.

## Testing (Windows)
Two helper scripts verify the feature end-to-end. Save them next to the repo root and run with `uv run python`.

```
uv run python demo_highs_plugin_setup.py hide     # copies bundled DLLs to custom_solver/, hides originals
uv run python demo_bundled_highs.py               # sets RTCTOOLS_EXTRA_CASADIPATH, runs pumped-hydro example
uv run python demo_highs_plugin_setup.py restore  # put originals back (also done automatically by demo_bundled_highs.py)
```

`demo_bundled_highs.py` hides the bundled DLLs, solves the pumped hydropower example via `RTCTOOLS_EXTRA_CASADIPATH=custom_solver/`, and restores the originals in a `finally` block.

Expected log line:
```
INFO ... Setting CasADi plugin search path to '<repo>/custom_solver;...'
```

<details>
<summary>demo_highs_plugin_setup.py</summary>

```python
"""
Helper script to temporarily rename the CasADi bundled HiGHS plugin so that
the custom solver in custom_solver/ takes effect when running demo_highs_plugin.py.

Two modes of operation:

  1. custom_solver/ already exists with DLLs:
     hide    — rename bundled DLLs to .bak so the custom ones win
     restore — rename .bak files back to their original names

  2. custom_solver/ does not exist:
     hide    — copy bundled DLLs into custom_solver/, then rename them to .bak
     restore — rename .bak files back (custom_solver/ stays in place)

Usage
-----
    uv run python demo_highs_plugin_setup.py hide     # rename bundled plugin away
    uv run python demo_highs_plugin_setup.py restore  # put the name back
"""

import shutil
import sys
import casadi
from pathlib import Path

REPO_ROOT   = Path(__file__).parent
CASADI_DIR  = Path(casadi.__file__).parent
CUSTOM_DIR  = REPO_ROOT / "custom_solver"

DLLS = [
    "libcasadi_conic_highs.dll",
    "libhighs.dll",
]


def hide():
    # If custom_solver/ does not exist, populate it from the bundled copies first.
    if not CUSTOM_DIR.exists():
        CUSTOM_DIR.mkdir()
        for name in DLLS:
            src = CASADI_DIR / name
            if src.exists():
                shutil.copy2(src, CUSTOM_DIR / name)
                print(f"Copied: {src} -> {CUSTOM_DIR / name}")
            else:
                print(f"Warning: bundled {name} not found at {src}, skipping copy")

    # Rename bundled DLLs to .bak so they are no longer found by CasADi.
    hidden_any = False
    for name in DLLS:
        bundled = CASADI_DIR / name
        hidden  = CASADI_DIR / (name + ".bak")
        if hidden.exists():
            print(f"Already hidden: {hidden.name}")
            continue
        if not bundled.exists():
            print(f"Not found, nothing to hide: {bundled.name}")
            continue
        bundled.rename(hidden)
        print(f"Renamed: {bundled.name} -> {hidden.name}")
        hidden_any = True

    if not hidden_any:
        print("Nothing new was hidden.")


def restore():
    restored_any = False
    for name in DLLS:
        bundled = CASADI_DIR / name
        hidden  = CASADI_DIR / (name + ".bak")
        if bundled.exists():
            print(f"Already restored: {bundled.name}")
            continue
        if not hidden.exists():
            print(f"Not found, nothing to restore: {hidden.name}")
            continue
        hidden.rename(bundled)
        print(f"Restored: {hidden.name} -> {bundled.name}")
        restored_any = True

    if not restored_any:
        print("Nothing to restore.")


if __name__ == "__main__":
    commands = {"hide": hide, "restore": restore}
    if len(sys.argv) != 2 or sys.argv[1] not in commands:
        print(f"Usage: {sys.argv[0]} hide|restore")
        sys.exit(1)
    commands[sys.argv[1]]()
```

</details>

<details>
<summary>demo_bundled_highs.py</summary>

```python
"""
Demo: RTCTOOLS_EXTRA_CASADIPATH with the same HiGHS version as bundled with CasADi.

Verifies that RTCTOOLS_EXTRA_CASADIPATH works correctly when the bundled DLLs are
not present in the CasADi folder, loading them instead from custom_solver/.

The script hides the bundled DLLs, runs the example via RTCTOOLS_EXTRA_CASADIPATH,
and restores the bundled DLLs afterwards.

Prerequisites
-------------
Run once to populate custom_solver/ if it does not exist yet:

    uv run python demo_highs_plugin_setup.py hide
    uv run python demo_highs_plugin_setup.py restore

Usage
-----
    uv run python demo_bundled_highs.py
"""

import os
import sys
from pathlib import Path
from unittest.mock import patch

REPO_ROOT   = Path(__file__).parent
EXAMPLE_DIR = REPO_ROOT / "examples" / "pumped_hydropower_system"
CUSTOM_DIR  = REPO_ROOT / "custom_solver"

if not CUSTOM_DIR.exists():
    print("ERROR: custom_solver/ not found. Run 'uv run python demo_highs_plugin_setup.py hide' first to populate it.")
    sys.exit(1)

# Set PATH and RTCTOOLS_EXTRA_CASADIPATH before importing casadi.
os.environ["RTCTOOLS_EXTRA_CASADIPATH"] = str(CUSTOM_DIR)
os.environ["PATH"] = str(CUSTOM_DIR) + os.pathsep + os.environ.get("PATH", "")

import demo_highs_plugin_setup
demo_highs_plugin_setup.hide()

sys.path.insert(0, str(EXAMPLE_DIR / "src"))

from rtctools.util import run_optimization_problem

with patch("rtctools.util.run_optimization_problem"):
    from example import PumpStorage

try:
    run_optimization_problem(PumpStorage, base_folder=str(EXAMPLE_DIR))
finally:
    demo_highs_plugin_setup.restore()
```

</details>

## Follow-up
- [CasADi #4299](https://github.com/casadi/casadi/pull/4299): Update bundled HiGHS version. Once merged, users can extract the updated plugin wrapper from a newer CasADi build and use it with this feature.
- [CasADi #4340](https://github.com/casadi/casadi/issues/4340): Switch to `LoadLibraryEx` on Windows — once merged, the transitive solver DLL will resolve without needing `PATH` manipulation.
- Integration test: add a CI-based test once a plugin wrapper build is available in the pipeline.

### DIFF
--- a/src/rtctools/util.py
+++ b/src/rtctools/util.py
@@ -26,6 +26,30 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
         return os.path.join(base_folder, subfolder)
 
 
+def _configure_extra_casadi_path(logger):
+    """Prepend RTCTOOLS_EXTRA_CASADIPATH to CasADi's plugin search path."""
+    if extra_casadi_path := os.getenv("RTCTOOLS_EXTRA_CASADIPATH", "").strip():
+        current_parts = [
+            p for p in (casadi.GlobalOptions.getCasadiPath() or "").split(os.pathsep) if p
+        ]
+        current_set = {os.path.normpath(p) for p in current_parts}
+        seen = set()
+        new_parts = []
+        for p in extra_casadi_path.split(os.pathsep):
+            if p and (norm := os.path.normpath(p)) not in current_set and norm not in seen:
+                new_parts.append(p)
+                seen.add(norm)
+        for p in new_parts:
+            if not os.path.isdir(p):
+                logger.warning(
+                    "RTCTOOLS_EXTRA_CASADIPATH entry %r does not exist or is not a directory.", p
+                )
+        if new_parts:
+            combined = os.pathsep.join(new_parts + current_parts)
+            logger.debug("Setting CasADi plugin search path to '%s'.", combined)
+            casadi.GlobalOptions.setCasadiPath(combined)
+
+
 def run_optimization_problem(
     optimization_problem_class, base_folder="..", log_level=logging.INFO, profile=False, **kwargs
 ):
@@ -99,6 +123,8 @@ def run_optimization_problem(
     # Log version info
     logger.info(f"Using RTC-Tools {__version__}.")
     logger.debug(f"Using CasADi {casadi.__version__}.")
+
+    _configure_extra_casadi_path(logger)
 
     # Check for some common mistakes in inheritance order
     suggested_order = OrderedSet(
@@ -231,6 +257,8 @@ def run_simulation_problem(
 
     logger.info(f"Using RTC-Tools {__version__}")
     logger.debug(f"Using CasADi {casadi.__version__}.")
+
+    _configure_extra_casadi_path(logger)
 
     # Run
     prob = simulation_problem_class(
--- a/src/rtctools/util.py
+++ b/src/rtctools/util.py
@@ -26,6 +26,29 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
         return os.path.join(base_folder, subfolder)
 
 
+def _configure_extra_casadi_path(logger):
+    """Prepend RTCTOOLS_EXTRA_CASADIPATH to CasADi's plugin search path."""
+    if extra_casadi_path := os.getenv("RTCTOOLS_EXTRA_CASADIPATH", "").strip():
+        current_path = casadi.GlobalOptions.getCasadiPath()
+        current_parts = current_path.split(os.pathsep) if current_path else []
+        seen = {os.path.normcase(os.path.normpath(p)) for p in current_parts}
+        new_parts = []
+        for p in extra_casadi_path.split(os.pathsep):
+            norm = os.path.normcase(os.path.normpath(p))
+            if norm not in seen:
+                seen.add(norm)
+                new_parts.append(p)
+                if not os.path.isdir(norm):
+                    logger.warning(
+                        f"RTCTOOLS_EXTRA_CASADIPATH entry {p!r} does not exist"
+                        " or is not a directory."
+                    )
+        if new_parts:
+            combined = os.pathsep.join(new_parts + current_parts)
+            logger.info(f"Setting CasADi plugin search path to '{combined}'.")
+            casadi.GlobalOptions.setCasadiPath(combined)
+
+
 def run_optimization_problem(
     optimization_problem_class, base_folder="..", log_level=logging.INFO, profile=False, **kwargs
 ):
@@ -99,6 +122,8 @@ def run_optimization_problem(
     # Log version info
     logger.info(f"Using RTC-Tools {__version__}.")
     logger.debug(f"Using CasADi {casadi.__version__}.")
+
+    _configure_extra_casadi_path(logger)
 
     # Check for some common mistakes in inheritance order
     suggested_order = OrderedSet(
@@ -231,6 +256,8 @@ def run_simulation_problem(
 
     logger.info(f"Using RTC-Tools {__version__}")
     logger.debug(f"Using CasADi {casadi.__version__}.")
+
+    _configure_extra_casadi_path(logger)
 
     # Run
     prob = simulation_problem_class(
--- a/tests/optimization/test_extra_casadipath.py
+++ b/tests/optimization/test_extra_casadipath.py
@@ -1,0 +1,125 @@
+import logging
+import os
+from unittest.mock import patch
+
+from rtctools.util import _configure_extra_casadi_path
+
+from ..test_case import TestCase
+
+logger = logging.getLogger("rtctools")
+
+
+class TestExtraCasadiPath(TestCase):
+    @patch.dict(os.environ, {"RTCTOOLS_EXTRA_CASADIPATH": "/custom/path"})
+    @patch("rtctools.util.casadi")
+    def test_prepends_to_existing_path(self, mock_casadi):
+        mock_casadi.GlobalOptions.getCasadiPath.return_value = "/casadi/default"
+
+        with self.assertLogs("rtctools", level=logging.DEBUG) as cm:
+            _configure_extra_casadi_path(logger)
+
+        call_args = mock_casadi.GlobalOptions.setCasadiPath.call_args.args[0]
+        self.assertEqual(call_args.split(os.pathsep), ["/custom/path", "/casadi/default"])
+        # The DEBUG log must mention the resulting combined path.
+        # assertLogs output format is "LEVEL:logger:message".
+        self.assertTrue(
+            any(call_args in r and r.startswith("DEBUG:") for r in cm.output),
+            f"Expected combined path in DEBUG log output, got: {cm.output}",
+        )
+
+    @patch("rtctools.util.casadi")
+    def test_absent_env_var_does_nothing(self, mock_casadi):
+        os.environ.pop("RTCTOOLS_EXTRA_CASADIPATH", None)
+        with self.assertNoLogs("rtctools", level=logging.DEBUG):
+            _configure_extra_casadi_path(logger)
+        mock_casadi.GlobalOptions.getCasadiPath.assert_not_called()
+        mock_casadi.GlobalOptions.setCasadiPath.assert_not_called()
+
+    @patch("rtctools.util.casadi")
+    def test_empty_or_whitespace_env_var_does_nothing(self, mock_casadi):
+        """Empty and whitespace-only values must both be treated as absent."""
+        for value in ("", " "):
+            mock_casadi.reset_mock()
+            with patch.dict(os.environ, {"RTCTOOLS_EXTRA_CASADIPATH": value}):
+                with self.assertNoLogs("rtctools", level=logging.DEBUG):
+                    _configure_extra_casadi_path(logger)
+            mock_casadi.GlobalOptions.getCasadiPath.assert_not_called()
+            mock_casadi.GlobalOptions.setCasadiPath.assert_not_called()
+
+    @patch.dict(os.environ, {"RTCTOOLS_EXTRA_CASADIPATH": "/custom/path"})
+    @patch("rtctools.util.casadi")
+    def test_no_separator_artefacts_when_existing_path_absent(self, mock_casadi):
+        """getCasadiPath() returning None or "" must not introduce separator artefacts."""
+        for return_value in (None, ""):
+            mock_casadi.reset_mock()
+            mock_casadi.GlobalOptions.getCasadiPath.return_value = return_value
+
+            with self.assertLogs("rtctools", level=logging.DEBUG):
+                _configure_extra_casadi_path(logger)
+
+            call_args = mock_casadi.GlobalOptions.setCasadiPath.call_args.args[0]
+            self.assertEqual(
+                call_args, "/custom/path", f"Failed for getCasadiPath()={return_value!r}"
+            )
+
+    @patch.dict(os.environ, {"RTCTOOLS_EXTRA_CASADIPATH": "/custom/path"})
+    @patch("rtctools.util.casadi")
+    def test_already_in_existing_path_is_not_prepended(self, mock_casadi):
+        """A path already present in the CasADi path must not be prepended again."""
+        mock_casadi.GlobalOptions.getCasadiPath.return_value = "/custom/path"
+
+        # No new paths to add, so setCasadiPath must not be called and no log emitted.
+        with self.assertNoLogs("rtctools", level=logging.DEBUG):
+            _configure_extra_casadi_path(logger)
+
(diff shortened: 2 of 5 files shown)


jarsarasty:
Hi @jackvreeken , @FaridAlavi 

The PR is ready for review! The manual test on Windows was successful!

jackvreeken:
@jarsarasty added a commit on top with some changes . See if you understand, and then you can squash them into your commit.

jarsarasty:
> @jarsarasty added a commit on top with some changes . See if you understand, and then you can squash them into your commit.

Thanks for the fix, @jackvreeken! I understand the changes well.

One thing I noticed: `test_empty_env_var_does_nothing` replaced `test_no_env_var_does_nothing` `(clear=True)`, but these cover different user-facing scenarios — the variable being absent vs. explicitly set to empty. They happen to behave the same today, but I think both are worth keeping as explicit documentation of intent. I'd also like to restore the `getCasadiPath.assert_not_called()` assertion that was dropped.

I will add those back after squashing.

src/rtctools/util.py
```diff
@@ -26,6 +26,17 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
...
 
+def _configure_extra_casadi_path():
+    """Prepend RTCTOOLS_EXTRA_CASADIPATH to CasADi's plugin search path."""
+    if extra_casadi_path := os.getenv("RTCTOOLS_EXTRA_CASADIPATH", "").strip():
+        current_parts = [
+            p for p in (casadi.GlobalOptions.getCasadiPath() or "").split(os.pathsep) if p
+        ]
+        new_parts = [p for p in extra_casadi_path.split(os.pathsep) if p and p not in current_parts]
+        if new_parts:
+            casadi.GlobalOptions.setCasadiPath(os.pathsep.join(new_parts + current_parts))
```
jackvreeken:
```Python
if new_parts:
              combined = os.pathsep.join(new_parts + current_parts)
              logger.debug(f"Setting CasADi plugin search path to '{combined}'")
              casadi.GlobalOptions.setCasadiPath(combined)
```
Could be nice for debugging purposes (to quickly check if it works correctly). Could also then "caplog" test this.

jackvreeken:
@jarsarasty put a commit on top (you can squash it) for debug logging (and also asserting that in the test). Could be useful feedback to the user. If you don't like it, feel free to remove the commit.

jarsarasty:
> @jarsarasty put a commit on top (you can squash it) for debug logging (and also asserting that in the test). Could be useful feedback to the user. If you don't like it, feel free to remove the commit.

Thank you for these fixes, @jackvreeken.

I have also implemented a few other improvements: 
- Path deduplication now uses os.path.normpath — handles trailing-slash variants (e.g. /path/one and /path/one/ treated as the same)
- Dedup applied within new_parts itself, not just against the existing CasADi path — prevents duplicates when the env var contains the same path twice
- logger.warning() emitted for each path entry that doesn't exist on disk 
- consolidated and hardened tests

src/rtctools/util.py
```diff
@@ -26,6 +26,30 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
...
+                new_parts.append(p)
+                seen.add(norm)
+        for p in new_parts:
+            if not os.path.isdir(p):
+                logger.warning(
+                    "RTCTOOLS_EXTRA_CASADIPATH entry %r does not exist or is not a directory.", p
+                )
+        if new_parts:
+            combined = os.pathsep.join(new_parts + current_parts)
+            logger.debug("Setting CasADi plugin search path to '%s'.", combined)
```
jackvreeken, line 49: No %s please. RTC-Tools uses f-formatted strings where possible.

src/rtctools/util.py
```diff
@@ -26,6 +26,30 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
...
+    """Prepend RTCTOOLS_EXTRA_CASADIPATH to CasADi's plugin search path."""
+    if extra_casadi_path := os.getenv("RTCTOOLS_EXTRA_CASADIPATH", "").strip():
+        current_parts = [
+            p for p in (casadi.GlobalOptions.getCasadiPath() or "").split(os.pathsep) if p
+        ]
+        current_set = {os.path.normpath(p) for p in current_parts}
+        seen = set()
+        new_parts = []
+        for p in extra_casadi_path.split(os.pathsep):
+            if p and (norm := os.path.normpath(p)) not in current_set and norm not in seen:
```
jackvreeken, line 39: Please don't do `if p`. An empty entry is valid on Linux, and means current dir. `os.path.normpath("")` resolves to "."

jackvreeken, line 39: If you rename "current_set" to "seen", you don't need this double thing and walrus operator.

src/rtctools/util.py
```diff
@@ -26,6 +26,30 @@ def _resolve_folder(kwargs, base_folder, subfolder_kw, default):
...
+                new_parts.append(p)
+                seen.add(norm)
+        for p in new_parts:
+            if not os.path.isdir(p):
+                logger.warning(
+                    "RTCTOOLS_EXTRA_CASADIPATH entry %r does not exist or is not a directory.", p
+                )
+        if new_parts:
+            combined = os.pathsep.join(new_parts + current_parts)
```
jackvreeken, line 48: + on lists is ugly, but we do it so much that you can leave it for now.

tests/optimization/test_extra_casadipath.py
```diff
@@ -0,0 +1,125 @@
...
+                    _configure_extra_casadi_path(logger)
+            mock_casadi.GlobalOptions.getCasadiPath.assert_not_called()
+            mock_casadi.GlobalOptions.setCasadiPath.assert_not_called()
+
+    @patch.dict(os.environ, {"RTCTOOLS_EXTRA_CASADIPATH": "/custom/path"})
+    @patch("rtctools.util.casadi")
+    def test_no_separator_artefacts_when_existing_path_absent(self, mock_casadi):
+        """getCasadiPath() returning None or "" must not introduce separator artefacts."""
+        for return_value in (None, ""):
+            mock_casadi.reset_mock()
```
jackvreeken, line 54: This is really rather ugly, but OK to leave for now. Because we're still "generic" in our test setup, we can't rely on pytest's parameterize; but if/when we move to pytest-formatted tests, I suggest we do that instead of this ugly explicit reset.

jackvreeken (changes requested):

FaridAlavi (commented):
Please consider updating the log messages. In the current implementation, if the user sets the environment variable, but for some reason the new solver file cannot be used by RTC-Tools, there is no log message indicating the error.

src/rtctools/util.py
jarsarasty: Implemented, and covered by `test_prepends_to_existing_path`. Thanks!

src/rtctools/util.py
jarsarasty: Done

src/rtctools/util.py
jarsarasty: Done. `current_parts = current_path.split(os.pathsep) if current_path else []` — preserves empty entries from the existing path; handles `None/""` from `getCasadiPath()`.

jarsarasty: Renamed

src/rtctools/util.py
jarsarasty: Left as-is for now.

tests/optimization/test_extra_casadipath.py
jarsarasty: I created an issue about future migration to Pytest: https://github.com/rtc-tools/rtc-tools/issues/1787

jarsarasty:
> Please consider updating the log messages. In the current implementation, if the user sets the environment variable, but for some reason the new solver file cannot be used by RTC-Tools, there is no log message indicating the error.

Hi @FaridAlavi 

Thank you for your feedback. The documentation has been updated, including now more detailed instructions about the path setup on Windows.

Regarding the log, missing-directory entries log a warning. Plugin load failures (ABI mismatch, missing transitive DLL) only surface at solver instantiation, where this function has no hook. Added a doc note pointing readers to the CasADi error in that case.

jarsarasty:
Thank you for the review, @jackvreeken 
Your suggestions have been incorporated: Debug log added; %s → f-string; if p removed; current_set renamed to seen.

The PR is ready for another look when you have time.

FaridAlavi (changes requested):
Currently, the pull request is not ready for merging. The main reason is that it has not been tested, and it is unclear for which solver versions this feature can be used—if it can be used at all. Additionally, there should be a solid justification for simply replacing the solver’s DLL, especially since there is no guarantee that the solver interface on the CasADi side is compatible with the new DLL file. In my opinion, the more reliable way to use newer versions of a solver is to compile CasADi specifically for that solver.

jarsarasty:
Hi @FaridAlavi 

Thank you for your comment.

The scope of this PR is to allow users to load a custom solver plugin via `RTCTOOLS_EXTRA_CASADIPATH`, for cases where a different HiGHS version is needed without recompiling all of CasADi. This scope was defined in #1735. 

As mentioned in the documentation, the plugin wrapper and solver library must come from the same HiGHS build to ensure ABI compatibility. This means the feature works with any HiGHS version, as long as the wrapper is compiled against the matching HiGHS headers.

To make building a compatible plugin straightforward, I am preparing a follow-up PR with a standalone CMake that compiles the plugin wrapper against the installed CasADi, without requiring a full CasADi recompile, which could be cumbersome.

For testing, here's the script we've been using locally — it runs the pumped hydropower example against a custom solver placed in custom_solver/:

```
import os
import sys
from pathlib import Path

EXAMPLE_DIR = Path(__file__).parent / "examples" / "pumped_hydropower_system"
CUSTOM_SOLVER_DIR = Path(__file__).parent / "custom_solver"

# Alternatively, set these in the shell before running:
#   set RTCTOOLS_EXTRA_CASADIPATH=custom_solver
#   set PATH=custom_solver;%PATH%
os.environ["RTCTOOLS_EXTRA_CASADIPATH"] = str(CUSTOM_SOLVER_DIR)
os.environ["PATH"] = str(CUSTOM_SOLVER_DIR) + os.pathsep + os.environ.get("PATH", "")

sys.path.insert(0, str(EXAMPLE_DIR / "src"))

from example import PumpStorage
from rtctools.util import run_optimization_problem

run_optimization_problem(PumpStorage, base_folder=str(EXAMPLE_DIR))  

```
I'm happy to follow up on the discussion at the office next week.

FaridAlavi:
Thanks @jarsarasty for the detailed implementation and clarification.

I am OK with merging this. One important point: since this allows overriding CasADi’s bundled plugins, it can affect solver behavior in ways that are environment-dependent and not always obvious. To make this manageable, it is important that this is clearly visible at runtime. Specifically, a clear log line when RTCTOOLS_EXTRA_CASADIPATH is used and overrides the default CasADi plugins helps debugging future RTC-Tools models of clients easier.

With that in place, I am happy to approve.

jarsarasty:
Hi @FaridAlavi, thanks for the review. I've addressed your comments:

- Promoted the log message from debug to info so users can confirm the override is in effect without enabling debug logging
- Included some additional clarifications in the documentation.
- Update the PR description with the test scripts
 
Let me know if anything else needs attention.

FaridAlavi (approved):

jarsarasty:
Thank you, @FaridAlavi

jarsarasty:
Hi @jackvreeken ,

All comments have been addressed.
Could you please have a look again and approve if everything looks good?

jackvreeken (approved):

PR #1780, opened 2026-04-08
test(modelica_mixin): replace repr() comparisons with numerical evaluation

### LINKED ISSUE #1759: Fix skipped test: CasADi sumRows()
tests/optimization/test_modelica_mixin.py:419 is skipped with "This test fails, because we use CasADi sumRows() now."
A failing test was skipped instead of fixed. Investigate what changed and update the test or implementation accordingly.

### LINKED ISSUE #1760: Fix skipped modelica mixin test with no reason
tests/optimization/test_modelica_mixin.py:388 is skipped with reason "Skipped" — no explanation. Investigate why and either fix it or document the reason properly.

jarsarasty:
`test_integral` and `test_states_in` were skipped because they compared
repr() of CasADi symbolic expressions, which is fragile — the symbolic
graph structure depends on implementation details. When integral()
switched to `ca.sum1()` (sumRows internally), the repr() no longer matched
the manually accumulated expression, even though both are mathematically
equivalent.

Both tests now evaluate expressions numerically at the optimal solution
via `ca.Function(solver_input)(solver_output)` and compare values with a
tolerance, decoupling the tests from CasADi's internal graph representation.

In `test_states_in`, each sub-case also asserts that `states_in()` returns
the expected number of points to produce a meaningful diagnostic on
boundary inclusion bugs instead of a criptic NumPy error.

  Closes #1759, #1760

### DIFF
--- a/tests/optimization/test_modelica_mixin.py
+++ b/tests/optimization/test_modelica_mixin.py
@@ -2,6 +2,7 @@ import logging
 import sys
 import unittest
 
+import casadi as ca
 import numpy as np
 from casadi import MX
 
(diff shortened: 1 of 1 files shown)


FaridAlavi (approved):
Nice improvement. One thing I noticed is that the off-grid `test_integral` cases are still not fully independent, since both `integral()` and the test-side trapezoid calculation rely on `state_at()` for interpolation. That means they mainly validate the interval assembly and accumulation path rather than interpolation itself. Since interpolation is already tested elsewhere, this looks acceptable to me.

jarsarasty:
> Nice improvement. One thing I noticed is that the off-grid `test_integral` cases are still not fully independent, since both `integral()` and the test-side trapezoid calculation rely on `state_at()` for interpolation. That means they mainly validate the interval assembly and accumulation path rather than interpolation itself. Since interpolation is already tested elsewhere, this looks acceptable to me.

Thank you, @FaridAlavi

PR #1788, opened 2026-05-11
ci: allow SonarQube scan step to fail without blocking CI

jarsarasty:
After renabling the SonarQube workflow, the CI is failing due to a missing SONAR_TOKEN.. I made a requiest to the LF to set up the new SONAR_TOKEN toean. 
As a temporary measure until the token is regenerated for the new rtc-tools-org SonarCloud organization, we can add `continue-on-error: true` to the SonarQube Cloud Scan step so that an invalid or missing `SONAR_TOKEN` does not fail the coverage job.


FaridAlavi (approved):
I created the Issue #1789 , to make sure we do not forget to revert this change.

PR #1803, opened 2026-06-15
fix(optimization): correct downhill logic and orifice flow constraint…

k-horvath-deltares:
…s in pumps and orifices example

In the mixed-integer example the pump was pumping the same time as the orifice was discharging. This is very rarely happening in practice, as pumps usually pump only uphill. A constraint is added to enforce this behaviour.

### DIFF
--- a/examples/mixed_integer/src/example.py
+++ b/examples/mixed_integer/src/example.py
@@ -27,16 +27,21 @@ class Example(CSVMixin, ModelicaMixin, CollocatedIntegratedOptimizationProblem):
     def path_constraints(self, ensemble_member):
         # Call super to get default constraints
         constraints = super().path_constraints(ensemble_member)
-        M = 2  # The so-called "big-M"
 
+        M = 10.0  # Big M, maximum possible discharge through the orifice
         # Release through orifice downhill only. This constraint enforces the
         # fact that water only flows downhill.
         constraints.append(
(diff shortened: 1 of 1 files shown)


examples/mixed_integer/src/example.py
```diff
@@ -35,6 +35,10 @@ def path_constraints(self, ensemble_member):
             (self.state("Q_orifice") + (1 - self.state("is_downhill")) * 10, 0.0, 10.0)
         )
 
+        # Pump upnhill only. This constraint enforces the
+        # fact that most pumps can only pump uphill.
+        constraints.append((self.state("Q_pump") + (self.state("is_downhill")) * 10, 0.0, 10.0))
+
```
FaridAlavi, line 41: Rather than using the number 10 directly, I suggest introducing a variable M with a value of 10. This makes it clear that a big-M constraint is being used. It is also helpful to explain that 10 is an appropriate value because the upper bound of Q_pump in the Modelica model is set to 7.

It also makes sense to apply a similar change to the previous constraint related to orifice as well (line 35).

FaridAlavi (changes requested):

examples/mixed_integer/src/example.py
k-horvath-deltares: I fixed it, thank you.

FaridAlavi (approved):

PR #1804, opened 2026-06-15
feat(simulation): add logging when state variable is initialized to zero

k-horvath-deltares:
Adds an informational log message during simulation to indicate when a state variable is initialized to zero. This helps improve transparency and traceability of model initialization, making it easier to diagnose unexpected behavior related to initial conditions.


FaridAlavi (commented):
The implementation is correct. However, SonarCloud raises a false warning about the comparison of floating-point numbers.

Please consider applying one of the following solutions:

1. Replace the equality check with something like `if abs(numeric_start_val) < epsilon:`
2. Find a way to disable the warning for this line. I am not sure if this is possible, but it is worth investigating.

jarsarasty:
> The implementation is correct. However, SonarCloud raises a false warning about the comparison of floating-point numbers.
> 
> Please consider applying one of the following solutions:
> 
> 1. Replace the equality check with something like `if abs(numeric_start_val) < epsilon:`
> 2. Find a way to disable the warning for this line. I am not sure if this is possible, but it is worth investigating.

@k-horvath-deltares 
You can use NOSONAR to disable this warning.

FaridAlavi (approved):

k-horvath-deltares:
Thank you for the comments and help, I suppressed the sonarcube check for that line.

src/rtctools/simulation/simulation_problem.py
```diff
@@ -508,6 +508,10 @@ def initialize(self, config_file=None):
                 var.min = start_expr
                 var.max = start_expr
             else:
+                if numeric_start_val == 0.0:  # NOSONAR - safe comparison due to settings above
+                    logger.info(
+                        f"Initialize: {var_name} will set to zero as no start value was found"
```
jarsarasty, line 513:
```suggestion
                        f"Initialize: {var_name} will be set to zero as no start value was found."
```

jarsarasty:
I just squashed the 4 commits into one, simplified the commit message, and fixed the log message wording. Since I did these changes, we need again an approval of this PR.

@FaridAlavi, could you please review again and approve?

FaridAlavi (approved):

jarsarasty:
Hi @k-horvath-deltares , could you please squash the commits on this PR into a single one? You can do it with:
```
git fetch origin
git reset --soft origin/master
git commit -s -m "feat(simulation): log when state variable is initialized to zero..."
git push --force-with-lease
```

FaridAlavi (approved):

BernhardBecker (approved):

BernhardBecker (approved):

PR #1805, opened 2026-06-16
fix(rtctoolsapp): download examples from release zipball

### LINKED ISSUE #1799: rtc-tools-download-examples fails with StopIteration due to stale org name prefix
The download_examples script downloads the repo zip from GitHub and looks for a folder starting with Deltares-rtc-tools.

Since the repo moved from the Deltares org to rtc-tools, the zip now contains a folder named rtc-tools-rtc-tools-<sha>/, causing StopIteration.

FaridAlavi:
Resolve the GitHub zipball version to the nearest release tag before downloading examples. This avoids failures for development versions such as 2.7.2+9.g0f23240, where no zipball is published. Also log the resolved release version used for the download.

### DIFF
--- a/src/rtctools/rtctoolsapp.py
+++ b/src/rtctools/rtctoolsapp.py
@@ -1,12 +1,34 @@
 import importlib.resources
 import logging
 import os
+import re
 import shutil
 import sys
 from importlib import metadata as importlib_metadata
 from pathlib import Path
 
 import rtctools
+
+
+def _zipball_version(version: str) -> str:
+    """Return the nearest GitHub zipball version for an RTC-Tools version string."""
+
+    version = version.split("+", 1)[0]
+
+    match = re.match(
+        r"^(?P<base>\d+(?:\.\d+)*)(?:(?P<pre>a|b|rc)(?P<pre_num>\d+))?(?:\.post\d+)?(?:\.dev\d+)?$",
+        version,
+    )
+    if not match:
+        return version
+
+    release_version = match.group("base")
+    pre = match.group("pre")
+    if pre is not None:
+        release_version += f"{pre}{match.group('pre_num')}"
+
+    return release_version
+
 
 logging.basicConfig(format="%(asctime)s %(levelname)s %(message)s")
 logger = logging.getLogger("rtctools")
@@ -72,7 +94,7 @@ def copy_libraries(*args):
         except OSError:
             sys.exit(f"Failed merging the libraries in package '{tld}'")
 
-    sys.exit(f"Succesfully copied all library folders to '{dst.resolve()}'")
+    sys.exit(f"Successfully copied all library folders to '{dst.resolve()}'")
 
 
 def download_examples(*args):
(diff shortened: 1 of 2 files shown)


src/rtctools/rtctoolsapp.py
```diff
@@ -94,28 +114,47 @@ def download_examples(*args):
...
-
-        sys.exit(f"Succesfully downloaded the RTC-Tools examples to '{target.resolve()}'")
+        sys.exit(f"Could not find examples for RTC-Tools version {release_version}.")
 
     try:
-        os.remove(local_filename)
-    except OSError:
-        pass
+        with ZipFile(local_filename, "r") as z:
+            import tempfile
```
jarsarasty, line 141: please move import tempfile alongside the other deferred imports at the top of the function

src/rtctools/rtctoolsapp.py
```diff
@@ -6,8 +6,28 @@
...
+    """Return the nearest GitHub zipball version for an RTC-Tools version string.
+
+    Raises:
+        InvalidVersion: If the version string is completely malformed.
+    """
+
+    v = Version(version)
+    release_version = ".".join(str(x) for x in v.release)
+
+    if v.is_prerelease and v.pre:
```
jarsarasty: dev versions (e.g. 2.8.0.dev5) will silently resolve to final release. Should probably rise a clear message like "Cannot download examples for a dev version; install a released version of RTC-Tools.

src/rtctools/rtctoolsapp.py
```diff
@@ -94,28 +114,47 @@ def download_examples(*args):
...
 
     try:
-        os.remove(local_filename)
-    except OSError:
-        pass
+        with ZipFile(local_filename, "r") as z:
+            import tempfile
+
+            target = path / "rtc-tools-examples"
+            zip_folder_name = next(
```
jarsarasty, line 144: This will not handle the case with a zip empty or malformed. You could use: 
```
zip_folder_name = next(
    (name.split("/", 1)[0] for name in z.namelist() if name and "/" in name),
    None,
)
if zip_folder_name is None:
    sys.exit("Downloaded archive is malformed or empty.")
```

tests/test_rtctoolsapp.py
```diff
@@ -0,0 +1,140 @@
...
+from pathlib import Path
+from unittest.mock import patch
+
+import rtctools
+from rtctools import rtctoolsapp
+
+
+class TestDownloadExamples(unittest.TestCase):
+    def test_download_examples_uses_release_version_and_logs(self):
+        with tempfile.TemporaryDirectory() as tmp:
```
jarsarasty, line 15: The three tests repeat ~25 lines (lines 14-38) of identical setup (temp dir creation, zip building, mock patches) with only the version string, zip folder name, and assertions varying between them. Extracting a shared helper method parameterised on those two inputs would make each test body much shorter and mean any future change to the setup only needs to be made once.

jarsarasty (changes requested):
Thank you, @FaridAlavi 
This PR fixes the reported issue. Additionally dev/local version strings no longer cause a 404, the zip extraction is org-agnostic, both typos are fixed, and tests are added.
However, I suggest some improvements related to handling dev versions, location of an import and code duplication.

src/rtctools/rtctoolsapp.py
FaridAlavi: In the function `download_example`, we already have:
```Python
if release_version != version:
        logger.info(
            "Using RTC-Tools version %s (resolved from %s) to download examples.",
            release_version,
            version,
        )
```

It means that for development versions, the log message shows exactly which version was requested and which version was downloaded.

src/rtctools/rtctoolsapp.py
jarsarasty: Hi @FaridAlavi This comment was marked as resolved, but the `import tempfile` is still inside the `with ZipFile` block, instead of at the top of `download_examples()` alongside `import urllib.request`. 

Not a blocker. We can merge, but next time if the change is not made, please explain why.

jarsarasty (approved):

src/rtctools/rtctoolsapp.py
```diff
@@ -94,28 +116,51 @@ def download_examples(*args):
     from zipfile import ZipFile
 
     version = rtctools.__version__
+    release_version = _zipball_version(version)
+    if release_version != version:
+        logger.info(
+            "Using RTC-Tools version %s (resolved from %s) to download examples.",
```
jarsarasty: I missed this one: we use f-strings for logging. 
We should update the CONTRIBUTING.md with all the conventions.

src/rtctools/rtctoolsapp.py
```diff
@@ -72,7 +94,7 @@ def _copytree(src, dst, symlinks=False, ignore=None):
         except OSError:
             sys.exit(f"Failed merging the libraries in package '{tld}'")
 
-    sys.exit(f"Succesfully copied all library folders to '{dst.resolve()}'")
+    sys.exit(f"Successfully copied all library folders to '{dst.resolve()}'")
```
jarsarasty, line 97: We use f-strings for logging.

src/rtctools/rtctoolsapp.py
jarsarasty: We use f-strings for logging.

src/rtctools/rtctoolsapp.py
```diff
@@ -94,28 +116,51 @@ def download_examples(*args):
...
     version = rtctools.__version__
+    release_version = _zipball_version(version)
+    if release_version != version:
+        logger.info(
+            "Using RTC-Tools version %s (resolved from %s) to download examples.",
+            release_version,
+            version,
+        )
+    else:
+        logger.info("Using RTC-Tools version %s to download examples.", release_version)
```
jarsarasty: We use f-strings.

src/rtctools/rtctoolsapp.py
jarsarasty: We use f-strings.

src/rtctools/rtctoolsapp.py
```diff
@@ -94,28 +116,51 @@ def download_examples(*args):
     from zipfile import ZipFile
 
     version = rtctools.__version__
+    release_version = _zipball_version(version)
+    if release_version != version:
+        logger.info(
```
jarsarasty, line 121: We use f-strings

jarsarasty (changes requested):
Hi @FaridAlavi 
The PR looks good, the examples can be downloaded, and the edge cases are covered. Please fix all the occurrences of the %-formatting — we use f-strings in the project.

src/rtctools/rtctoolsapp.py
FaridAlavi: Let's update CONTRIBUTING.md file in a separate pull request.

jarsarasty (approved):

jarsarasty:
Thank you, @FaridAlavi

PR #1824, opened 2026-09-04
CollInt: Fix nested delays for 'inline_delay_expressions'

jackvreeken:
**Response by Claude:**

Nested delays evaluate their inner delay as zero under `inline_delay_expressions` (option added in #1712), because `ca.substitute()` does not re-scan what it inserts. Mechanism and fix are in the commit message.

**Test.** `TestNestedDelay` in `tests/optimization/test_delay_expressions.py` uses a new `ModelNestedDelay.mo` with `y = 2 * delay(delay(u, 0.1), 0.1)` on a known constant input and checks both modes against the input shifted by four steps. On master the inline mode fails with `y` reading as all zeros:

```
E           AssertionError: abs(a - b) >= tol.
E           array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
E                  0., 0., 0., 0.])
E           array([2. , 2. , 2. , 2. , 2. , 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8,
E                  2.9, 3. , 3.1, 3.2, 3.3, 3.4, 3.5, 3.6])
```

With the fix, `uv run pytest --ignore=tests/examples tests` gives 363 passed, 1 skipped. Also verified outside the test suite: the underscore-elimination form gives the same values in both modes, a circular model (`a = delay(b, dt); b = delay(a, dt)`) raises in inline mode and still solves in the default mode, and a nested delay with history before t0 agrees between the two modes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jbaayen (approved):

jackvreeken:
@jarsarasty can you take a look, and see if you agree? It's kind of blocking for us now. Essentially (see the code and the commit message) it's doing a loop over

```
                delayed_feedback_variable_replacement = ca.substitute(
                    delayed_feedback_variable_replacement, X, delayed_feedback_variable_replacement
                )
```

until there's no longer a delay expression that depends on another delay expression.

jarsarasty (approved):

jarsarasty:
Thanks for this fix, @jackvreeken! 
Approved and merged

PR #1817, opened 2026-08-13
docs(optimization): document initialization of distributed Saint-Venant models

k-horvath-deltares:
Add theory and examples describing initialization of distributed
Saint-Venant models used in optimisation. Explain the relationship
between degrees of freedom, boundary conditions, and steady-state
constraints, and compare full-state and steady-state initialization
approaches.

Include example cases illustrating how inconsistent initialization
can lead to aggressive control actions and unrealistic hydraulic
transients.

PR #1829, opened 2026-09-22
BSpline1D: Fix fit allocating k+1 unused coefficients

jackvreeken:
## Problem

The "Newest Dependencies" CI job fails on `TestBSpline1DFit.test_fit` since CasADi 3.8.1 (see the failures on #1817). The PR itself is docs-only; the failure is a latent bug in `BSpline1D.fit`.

`fit` allocated one NLP variable per knot, but a degree-k spline on n knots has only n − k − 1 basis functions, and `BSpline1D.__call__` reads exactly that many weights. The trailing k + 1 variables never entered the objective or the curvature constraints; they were held apart only by the monotonicity constraints and were otherwise free. Every fit was therefore underdetermined, and the values of those variables depended on where Ipopt happened to stop.

CasADi 3.8.1 evaluates the expanded expressions with a last-bit difference from 3.8.0. On this unscaled problem of 1e8 magnitude that is enough to change Ipopt's iteration path: 3.8.0 stalls at the rounding floor and exits "acceptable" after 14 iterations, 3.8.1 converges cleanly in 10. The free variables landed 6.5e-4 relative away from the values the test hard-codes.

## Fix

Allocate n − k − 1 coefficients so the problem has a unique solution, and drop the four reference values from the test that never corresponded to a real coefficient. The test keeps its original tight comparison.

The returned `tck` has a shorter coefficient array. `BSpline1D.__call__` and scipy's `splev` both accept exactly n − k − 1 weights, and `tck` tuples cached by the old code still evaluate because the extra weights are ignored.

## Verification

- Fitted coefficients identical to ten digits with CasADi 3.6.3, 3.6.7, 3.7.0, 3.7.2, 3.8.0 and 3.8.1.
- `tests/data/test_interpolate.py` passes on all six versions; before the fix it fails on 3.8.1.
- Full unit suite passes in the default environment; `lookup_table` and `ensemble` examples run with fresh spline caches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jarsarasty (approved):

jarsarasty:
Thank you, @jackvreeken!

tests/data/test_interpolate.py
```diff
@@ -71,10 +71,6 @@ def test_fit(self):
                 2.07664058e08,
                 2.57930573e08,
                 2.63110885e08,
-                2.63210885e08,
-                2.63310885e08,
-                2.63410885e08,
-                2.63510885e08,
```
jbaayen, line 77: To make this bulletproof, can you also extend the test case to test approximate equality when evaluating the fit over a list of trial points? The reference values would be obtained using the old code.

jbaayen (changes requested):

tests/data/test_interpolate.py
jackvreeken: **Response by Claude:**

Added. The test now evaluates the fitted spline at nine trial points, the five data points plus four in between, and compares against values produced by the old `fit` from master under CasADi 3.8.0. The old and new code agree at those points to a maximum relative difference of 3e-16 under CasADi 3.6.3, 3.8.0 and 3.8.1, so the curve is unchanged and only the unused trailing coefficients are gone.

jackvreeken: **Response by Claude:**

Trimmed to five trial points, one inside the support of each basis function, which is exactly what is needed to determine a five-coefficient spline. The reference values are still the old code's output, and old and new agree at those points to machine precision under CasADi 3.6.3, 3.8.0 and 3.8.1.

jbaayen (approved):

